PR #54, opened 2022-02-24
Dispute system v2

Reckless-Satoshi:
Redesign of the dispute system so it happens completely inside the platform.

- [x] In a first step the frontend of the dispute system is finished. And instructions clearly specify that dispute resolution communication with staff, in the current version, happens via alternative channels (burner contact method)

- [ ] The payout to dispute winner (escrow+bond) is computed and the dispute winner is asked for a new invoice in the contract box. Payout happens automatically just like any other buyer payouts.

- [ ] Two way chat with admin in-app. Requires admin chat pannel. Requires chat persistent memory. Probably best to do when revisiting the chat app.


Reckless-Satoshi:
Okay, this PR is becoming a mess as my attention has shifted away from the dispute system into other priorities (I keep adding unrelated things into this branch). Given that the commits here are all running in the mainnet platform I will merge as is. But there is still two important tasks pending to take the dispute system to the next level.

PR #59, opened 2022-03-02
Chatbox fixes

Reckless-Satoshi:
In this PR: 
1 - Recommend users to PGP encrypt their sensitive information in the chat box. 
2 - Facilitate long strings sharing in chat.
3 - Small fix that prevents sending empty messages.

As discussed in the main telegram group. Users should not trust the chat being encrypted nor it not being logged. Temporarily, the chat helper information now links to a small tutorial on how to setup OpenKeychain and use it for messaging. This one is to be replaced by: A tutorial explicitly made for OpenKeychain+RoboSats (#60 looking for contributors)

Previously the chat was protected for overflow by long strings by inserting a space-bar every 22 characters. This is very inconvenient when copy-pasting encrypted strings, since one has to manually remove the space-bars. The fix now allows for graceful overflow that does not insert new characters in the string.


Reckless-Satoshi:
It is very unlikely these fixes broke any previous feature. 
Will test it a bit and if everything is OK will merge and update the platform to commit https://github.com/Reckless-Satoshi/robosats/pull/59/commits/191a0433030c4d1797e1b7e779cfdff569370757 by tonight GMT.

PR #61, opened 2022-03-03
Upgrade fee structure

Reckless-Satoshi:
This PR upgrades the fee structure. So far it was very basic: the seller locks a hold invoice for X, the buyer submit an invoice of 0.998*X. The overall fee for a trade is 0.2%.

This is functional, however the fee load is fully applied to the buyer (...well the market can accommodate it by shifting the equilibrium premium). However, we would prefer to use a proper fee structure to incentivize liquidity providers to use the platform. 

**So the fee structure changes from:**

|   | Maker | Taker |
|---|-----|---|
| Buyer | 0.2% | 0.2% |
| Seller | 0.0% | 0.0% |

**To:**

|   | Maker | Taker |
|---|-----|---|
| Buyer | 0.025% | 0.175% |
| Seller | 0.025% | 0.175% |

**The total trade fee remains the same (0.2%)**, but now we align the maker/taker incentives to better fit the platform's need of makers.

Why exactly that split between maker/taker? This split is taken from the predominant BTC p2p market (Bisq): https://bisq.wiki/Trading_fees . The fees on Bisq are: 0.1% maker fee and 0.7% taker fee. There seems to have been a lot of deliberation around this split for Bisq (makers pay a fraction of 0.125 of the total fees), so it makes for a very a good initial value for RoboSats.

It all boils down to a new hparam: `MAKER_FEE_SPLIT=0.125`  (It is implicit that TAKER_FEE_SPLIT = 1 - MAKER_FEE_SPLIT)


Reckless-Satoshi:
Everything works as expected. Already deployed in RoboSats mainnet.

For the future, one can think of also biasing the cost in the buyer-seller axis. As of now, buyer's routing fee is covered by the platform, while the seller's fee is not. In addition, there is an obvious imbalance between buyers and sellers.

Issue #52, opened 2022-02-20
Advanced options for makers


Reckless-Satoshi:
The maker page is simple, yet it should be sufficient for most casual users. However, some advanced options can be added in an "Advanced Options" tab. These are mostly directed to fulfill the needs of larger liquidity providers and advanced users. 

### Possible advanced options
- **Set public expiry time.** A datetime input to explicitly specify the time of expiry of a public order. This input will have to be limited to a max of ~10 hours from 'now' in order to not have too long-lived HTLCs. Example usage:
  - A maker has only 30 minutes for lunch or before going to bed. He places an order with a expiry time 30 minutes in the future. If it is taken he will be available for the trade, else the order expires no action needed.
- **Set fiat exchange duration.** It has recently been bumped to 24 hours. I deem it already too long (probably unsafe). Yet, makers might still want to ask for very slow payment methods such as bank transfers or cash by mail (totally not advised!). This option could allow for CLTV expiry of up to 1 week. It will have to be extensively tested in testnet first. (EDIT: longer CLTVs will probably never happen, looks bad, see #56 for more info)
- **Amount range.** the maker enters the minimum and maximum order size. The taker will select the exact amount just after "Take Order" button is pressed. Challenge: it is harder to compute a fair maker bond size.
- **Set bond size.** Bounded 1-20%. Will be the size of both, maker and taker bonds.
- **Allow bond-less takers.** How is someone new to LN going to buy Satoshis if he needs to lock satoshis in the first place? Makers might choose to allow bond-less takers in exchange of collecting a higher premium for the extra counterpart risk.

Submit your comments and ideas so we can craft this feature the right way before putting the work into it.

### Challenges
Most of these options will make the overall UI messier in other pages. For example, the book and order page will need to accommodate the extra information of these advanced options so takers are aware of the details of the order. That might make new users dizzy. There might be a way to simply add a check in the order page "Show advanced", otherwise the advanced orders will not show.

decentralizedb:
> * **Allow bond-less takers.** How is someone new to LN going to buy Satoshis if he needs to lock satoshis in the first place? Makers might choose to allow bond-less takers in exchange of collecting a higher premium for the extra counterpart risk.

About the bond-less takers topic. I think this option should be had by default some limits, one of them does not exceed some sats amount, maybe the minimum amount allowed in the platform (20K for example.). Another one should be to have a premium minimum higher by default, 15% eg., but it could be increased by the Maker.

[...]

PR #76, opened 2022-03-15
Advanced maker options v1

Reckless-Satoshi:
See #52 
- [x] Add Advanced options switch and page
- [ ] Implement _Allow bondless takers_
- [x] Implement _Set public expiry time_
- [x] Implement _Amount range_
- [x] Implement _Set bond size_


Reckless-Satoshi:
While allowing bond-less takers has not been implemented (a bigger undertaking than previously considered), I will merge this branch. It is already running on the mainnet platform.

PR #77, opened 2022-03-17
Robosats Logo

MishoKanaria:
This is the Logo to Robosats made by Misho


Reckless-Satoshi:
These logos are genuinely cool. They are going to be featured in many places :)

PR #80, opened 2022-03-24
Update currencies.json

tradewind886:
interesting project. adding TWD (New Taiwan Dollar)


tradewind886:
from api/utils.py I found that yadio.io doesn't support TWD
TWD is supported in blockchain.info

How best to handle this?

Reckless-Satoshi:
Hey @tradewind886 ,

Thanks for adding new currencies! 

Indeed, to add a new one, we only need to introduce a new symbol on [currencies.json](https://github.com/Reckless-Satoshi/robosats/blob/main/frontend/static/assets/currencies.json) and a flag emoji in [getFlags.js](https://github.com/Reckless-Satoshi/robosats/blob/main/frontend/src/components/getFlags.js)

A new symbol will only work if the currency is supported by _**at least one of the APIs**_ we use for fetching prices. If TWD is supported by blockchain.info, it will work! However, would be best to have some level of redundancy.

It might be about time to add a third source for market prices. Any idea for another reliable and independent source of prices that does not require authentication?

tradewind886:
Thanks for replying and pointing me in the right direction on what else to
update!

I will reopen the pull request once I reach out to yadio.io to add TWD

I don't know of free/non-authenticated reliable source for exchange rates
at the moment

tradewind886:
@Reckless-Satoshi 
Yadio.io has now added TWD onto their site ! Should work now

Reckless-Satoshi:
> I will reopen the pull request once I reach out to yadio.io to add TWD

So you just reached to them and they added a new currency? Woah, nice :D

PR #82, opened 2022-03-24
Update currencies.json

tradewind886:
adding TWD (new Taiwan Dollar) currency


Reckless-Satoshi:
Tested and merged. TWD will be available for trading in the next mainnet platform update.

PR #83, opened 2022-03-26
Create Fiat Payment Guide

bkminer1984:
Create Fiat Payment Guide

This PR create detailed recommendation for buyer and seller in trading with their peers.

Although the robosats is in the early stage, it'll eventually bring in more people with less tech skill and also more scammers.
Fiat payment process in the process that have huge variations and thus, huge risk of fraudulent for scammers and risk of error for new users. By creating more detailed fiat payment guidance will make it easier to on boarding more user to the platform with lesser loss on errors and fraudulent activity.

Objective of recommendation

1. Help new users get their first trade successfully.
2. Minimize the loss caused by fraudulent users.
3. Minimize unnecessary dispute/bond slashing from users mistake.

List of recommendation

- General recommendation for both peers
Mainly recommends about negotiate and agreeing on payment method.

- Recommendation for buyer
Focus on the things they need to verify and record before/after fiat transaction.

- Recommendation for seller
Focus on the things they need to verify and record before/after fiat transaction.

- Recommendation for those who using high-chargeback-risk method (Now, there's only paypal)
Add the recommendation to use paypal friend to prevent chargeback risk.


Reckless-Satoshi:
Woah @bkminer1984 this is very cool and useful! 

Feels like this is a nice basis to start building a good-practice-guide for payment methods. It can be completed by smartly directing users to Bisq and Hodlhodl guides.

I would ask for a few fixes before merging this doc.

1. Stick with the naming convention of `/docs/`, in this case `fiat-payment-methods.md`
2. The text can benefit of grammar and spelling checker. If English is not your main language, you can work on this doc on your native language. We worry about translating later.
3. The wording and explanations about Paypal are not straightforward. Only someone who already understand the issue and the solution can understand the paragraph. "Paypal friend" is an ambiguous term, and readers might not know it is a Payment modality.  Please state the full name of the modality of payment and link to the right resources about that modality by Paypal. Otherwise, screenshots might also help.
4. Enumerations with proper markdown formatting ( add a space between the point and the sentence)
5. As a buyer, the 1st point in the list is confusing since there is nothing you should actually wait for. RoboSats will not allow you to exchange payment details before the trade collateral is locked :)

Thanks a lot! This work is veeery useful for many!

bkminer1984:
Thank you for the recommendation! I've been improved the doc following your comments:

1. Edited the file name to conform with the naming convention.
2. Use Grammarly to correct the grammar (I'm not native English but in this case, I create the doc in English so I think I'm better sticking with English for this doc.)
3. Thanks a lot for pointing that out, I try to add more details by explaining the issue that can happen and the potential solution available. However, this section will probably use more spaces and I think I'll need to spilt "High-risk method discussion" into separate doc files later.
4. Edited
5. Yeah, at first I thought it is possible to add payment details in the payment method section (if you are the maker) but I think it is my misunderstanding, as people are more incentivized to keep payment details secret and only directly tell their peer (not tell everyone by announcing it in the order book) so I agree to remove the first point from the doc.

I also added the link to bisq trading rules and bisq wiki for fiat payment methods in the introduction section of the doc.

Reckless-Satoshi:
@bkminer1984 
Amazing work! A very useful resource for RoboSats users no doubt, thanks a lot for your dedication.

PR #84, opened 2022-03-29
Add autocomplete, icons, custom selector

Reckless-Satoshi:
- [x] Add payment method icons
- [x] Add autocomplete maker component
- [x] Wire it all
- [x] Parse OrderBook payment text into icons


Reckless-Satoshi:
Seems to work super smoothly in RoboTest. Deploying to Mainnet as soon as tomorrow.

Issue #35, opened 2022-01-15
Multilanguage Support


Reckless-Satoshi:
Make RoboSats available in more than just English.

dbolser:
Does the code currently support i18n? e.g. is this just an issue of finding translations? Or does the i18n work need to be done on the code?

Reckless-Satoshi:
> Or does the i18n work need to be done on the code?

This has to be done first.

In addition. There are a few places where English sentences are composed via scripts that will be difficult to make work on other languages.

The UI and texts are still very fluid. Let's wait at least until v0.2.0 release before putting the work into internationalization.

Reckless-Satoshi:
Work in progress.

PR #85, opened 2022-04-02
Internationalization

Reckless-Satoshi:
Set up react i18next and locales to kick off translation work


Reckless-Satoshi:
Fixes #35 and #86

PR #94, opened 2022-04-14
Extract getCookie function into its own file

fernandoporazzi:
The `getCookie` function was being implemented a couple of times across the pages. The changes in this PR create a single implementation to retrieve a cookie value by name and imports the helper function whenever needed.


Reckless-Satoshi:
Hey @fernandoporazzi,

Indeed, this was a bad hack early on that survived until now. This is probably also the case for other functions, I can recall pn() also being copy/pasted into the different components.

Will give it a check and merge once I have some time.

Happy to see a JS developer checking over the frontend components.  Thanks a lot!

PR #98, opened 2022-04-17
BTC instead of Sats on Stats For Nerds

ibertario:
In the "Stats For Nerds" we have now:

image

With this pretty good amount of daily volume, is better to have both amounts denominated in BTC:

image

This PR tries to solve that. Can't test but only two minor changes.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -730,7 +730,7 @@ class InfoView(ListAPIView):
             lifetime_volume = 0
 
         context["last_day_nonkyc_btc_premium"] = round(avg_premium, 2)
-        context["last_day_volume"] = total_volume *100000000
+        context["last_day_volume"] = total_volume
         context["lifetime_volume"] = lifetime_volume
         context["lnd_version"] = get_lnd_version()
         context["robosats_running_commit_hash"] = get_commit_robosats()
--- a/frontend/src/components/BottomBar.js
+++ b/frontend/src/components/BottomBar.js
@@ -152,7 +152,7 @@ class BottomBar extends Component {
                 <Divider/>
                 <ListItem>
                     <ListItemIcon><EqualizerIcon/></ListItemIcon>
-                    <ListItemText primary={pn(this.state.last_day_volume)+" Sats"} secondary={t("24h contracted volume")}/>
+                    <ListItemText primary={pn(this.state.last_day_volume)+" BTC"} secondary={t("24h contracted volume")}/>
                 </ListItem>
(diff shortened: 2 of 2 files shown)


Reckless-Satoshi:
Awesome!! I have not tested, but it looks like this PR should work!

Are you sure about this upgrade, though? Daily amounts are always going to be much smaller than lifetime amounts. So it makes sense to have BTC for lifetime, yet it also makes sense to have Sats for days with low trade volumes (for example, a few days ago the trade volume was only 200.000 Sats).

I propose, we pick a unit (either BTC or Sats). And then, implement a clickable action, so when the unit is clicked, the UI rotates units and multiplies or divides by 100000000.

ibertario:
> Daily amounts are always going to be much smaller than lifetime amounts. So it makes sense to have BTC for lifetime, yet it also might be nice to have Sats for those days with low trade volumes (for example, a few days ago the trade volume was only 200.000 Sats).

Well, talking about stats, amounts with the same denomination looks better, especially when they are very close.

I think that for a Bitcoin amount BTC is easy to look (especially for Nerds).

> I propose, we pick a unit (either BTC or Sats). And then, implement a clickable action, so when the number is clicked, the UI rotates units and multiplies or divides by 100000000.

That's beyond my current dev capabilities, but definitely can be good to show a button to exchange or a tooltip with the other option.

ibertario:
_PS: A button present in all UI that exchange between both BTC<>Sats can be (crazy?) valuated for the future._

ibertario:
> Daily amounts are always going to be much smaller than lifetime amounts. So it makes sense to have BTC for lifetime, yet it also might be nice to have Sats for those days with low trade volumes (for example, a few days ago the trade volume was only 200.000 Sats).

Some options to valuate...

* Add 1 week
* Last 100 or 1000 blocks stats
* Little graph with 4 points from the start
* Choice one good % increase (trades, Robots, medium sats trade...)

ibertario:
This can be merged! As we can see 24h volume is increasing...

image

We can add later another metric:

> Some options to minimize the divergence...
> 
>     * Add 1 week
> 
>     * Last 100 or 1000 blocks stats
> 
>     * Little graph with 4 points from the start
> 
>     * Choice one good % increase (trades, Robots, medium sats trade...)

Reckless-Satoshi:
Merging as is. I am still looking forward to have some free time for a full revamp of the summarizing stats :)

PR #99, opened 2022-04-19
Pretty numbers

fernandoporazzi:
The function `pn()` is being implemented a couple of times in different files across the project.

This PR removes the code duplication by co-locating the `pn()` helper function inside the `utils` folders.

Jest is also being introduced for unit tests.


Reckless-Satoshi:
Looks good to me!🚀 Thanks for introducing Jest.

PR #100, opened 2022-04-19
Convert InfoDialog class into a functional component

fernandoporazzi:
This PR converts the InfoDialog component class into a functional component. This is better for maintainability and also reduces bundle size, therefore increasing the performance of the application.


Reckless-Satoshi:
This PR looks good. I will test it and merge in the following days.

The app is mostly based on class components with some functional components here and there. Class components are expected to be poorly maintained (...or deprecated) in the future, so it is needed to convert them into functional components eventually. Thanks a lot for converting the InfoDialog!

PR #102, opened 2022-04-21
Remove built main.js from versioning

fernandoporazzi:
We should never version built files. This leads to conflicts when merging branches.

This commit deletes the main.js file and changes the .gitignore to prevent it from being added again


Reckless-Satoshi:
LGTM, merging!

It was a hassle. Yet it made it easy to deploy on the production platform since "git pull" was enough (no need to have the dependencies needed to build it). I will figure out an alternative pipeline for deploying to production.

PR #108, opened 2022-04-24
de.json

dsitti:
work in progress


dsitti:
From my side, it´s now ready to marge in to the testnet for a final review.

Reckless-Satoshi:
Hey @dsitti 

It is up in [testnet](robotestagw3dcxmd66r4rgksb4nmmr43fh77bzn2ia2eucduyeafnyd.onion/)! :)

Similar comment to the other translations: 

There are some places where the new text breaks the UI. For example, the homepage buttons become double line in the mobile view (not too relevant, just an example). Also some headers of the contract box become double line (you can create a dummy order to check the contract box). 

Any translation that takes more lines can lead to some parts of the UI being no longer visible (specially in small screens). Maybe you find an easy way to make these shorter, since translations often do not need to be literal. For example "Make order" and "View book" buttons could simply be "make" and "take" if that makes the new language fit. Some strings might be impossible to shorten, that's okay :)

Once again, thank you very much for this valuable contribution!

Reckless-Satoshi:
@dsitti Looks great! It will be up in the mainnet platform in the next upgrade this weekend

PR #109, opened 2022-04-24
Add OpenPGP.js

Reckless-Satoshi:
Work In Progress.
Use OpenPGP.js primitives to do keypair generation, key exchange a encrypt/decrypt messages on the chat.js component.

- [ ] Generate privKey deterministically from user token (needs research). If not possible, use OpenPGP to generate a short primer of the private key that will become the new way of generating user private tokens (tokens-v2). However it happens, a user privkey should be linked to their robot avatar and none will be ever send on a request. There might not be any off-the-shelf solution to this.
- [ ] Exchange public keys via Chat WS.
- [ ] Always PGP encrypt every message before sendding them over chat WS. 
- [ ] On message received, use own privKey to convert the received message to plaintext.
- [ ] Store an array in app state all previous messages encrypted/decrypted. By default the chat will look just as it does now. The array must contain for all messages: `message_id, message_plaintext, message_encrypted, message_time`
- [ ] By clicking the "Audit" button, users will see all of the `message_encrypted` in the array (the chat will be full with many "---- BEGIN PGP----(...)"). While this button is not pressed, users only see the list of `message_plaintext`.
- [ ] Implement a Dialog pop up with the pieces required to independently audit the E2E encryption: 1) user token/primer, 2) user privKey (tin an easy to export format) and 3) counterpart pubKey. 
- [ ] Pop up dialog has a link to a doc on how to audit the RoboSats' chat e2e encryption using OpenKeychain or OpenPGP on CLI. The doc could be a simple extension on what already exists on https://learn.robosats.com/docs/pgp-encryption/
- [ ] ... Optional: now the backend can safely store the chat logs and serve them back once a robot connects to the chat room.


frontend/src/components/Chat.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { withTranslation, Trans} from "react-i18next";
 import {Button, Link, Badge, TextField, Grid, Container, Card, CardHeader, Paper, Avatar, FormHelperText, Typography} from "@mui/material";
 import ReconnectingWebSocket from 'reconnecting-websocket';
+import * as openpgp from 'openpgp/lightweight';
```
fernandoporazzi, line 5: I see this being imported, but it never gets used. Is it supposed to be like this?

Reckless-Satoshi: Sorry, this is work in progress. But somehow that commit also slipped trough bad "git checkout -b newbranch" into another feature branch that is already merged into master :)

Reckless-Satoshi:
After some more thought and plannig. This feature is currently being developed in PR https://github.com/Reckless-Satoshi/robosats/pull/147

PR #113, opened 2022-04-28
Add amazon icons including flags of 8 regions

ozoono:
New amazon icon in .png format, plus 8 more icons including country flags:
spain, france, germany, italy, uk, united states, canada and china


Reckless-Satoshi:
They look good! 

Merging, without having checked whether they print nicely when small :) We can re-upload new ones if needed. Thanks a lot!

PR #122, opened 2022-05-02
Update de.json

dsitti:
Update de.json with the current en.json in main.


Reckless-Satoshi:
@dsitti 

Great! Thank you for noticing the new lines introduced with the latest PRs. I will be merging as is.

PR #127, opened 2022-05-07
Fix order book alignement in iOS

ozoono:
Fix order book alignement in iOS, preventing it from appearing to the right (It would be necessary to check that this change doesn't affect Android devices)

### DIFF
--- a/frontend/static/css/index.css
+++ b/frontend/static/css/index.css
@@ -136,3 +136,9 @@ input[type=number] {
 }
 
 .MuiButton-textInherit {color : '#111111';}
+
+@media (max-width: 929px) {
+  .MuiDataGrid-columnHeaders + div {
+    width: auto !important;
+  }
+}


Reckless-Satoshi:
Thanks a lot @ozoono ! Works as expected, no side effects seen.

frontend/static/css/index.css
```diff
@@ -135,4 +135,10 @@ input[type=number] {
   filter: drop-shadow(0.5px 0.5px 0.5px #000000);
 }
 
-.MuiButton-textInherit {color : '#111111';}
\ No newline at end of file
+.MuiButton-textInherit {color : '#111111';}
+
+@media (max-width: 929px) {
```
Reckless-Satoshi, line 140: Hey @ozoono , now that the app is also available on Umbrel some users are finding again this book misalignment on Safari for MacOS. Do you think removing this `929px` limit might extend this .css fix for Safari for desktop as well?

ozoono: Hi, @Reckless-Satoshi i think moving the first css block of this media-query:
 ```
 .MuiDataGrid-columnHeaders + div {
    width: auto !important;
  }
```
outside the media-query (so it will affect mobile and desktop), should work.
I am going to do a PR with this small fix

ozoono: I've manually checked this change on Safari and it works.

PR #129, opened 2022-05-08
Begin Thai translation

bkminer1984:
This commit is to create how-to-use_th.md and frontend/src/locales/th.json to begin translation into Thai.
The translation hasn't finished for both files, this could open an opportunity if there are other Thais who want to contribute the translation to QC or continue on the other paths.


bkminer1984:
I'm closing this PR to Push another PR with completed translation later.
(It is completed but need some more grammar checking in thai)

PR #130, opened 2022-05-08
Fix layout for small iOS devices and allow scroll

ozoono:
Add css styles to fix the layout and allow scrolling in small iOS devices in order to solve 2 problems:
- top selectors are not totally visible
- bottom buttons are not visible (neither reachable because is not possible to scroll down) 

Added styles are only supported in Safari, so Android devices shouldn't be affected

### DIFF
--- a/frontend/static/css/index.css
+++ b/frontend/static/css/index.css
@@ -142,7 +142,7 @@ input[type=number] {
     width: auto !important;
   }
 
-  .appCenter:has(>div.MuiGrid-root:first-child) {
+  .appCenter:has(>div.MuiGrid-root:first-child, >div.MuiBox-root:first-child) {
     overflow-y: scroll;
     margin-top: 12px;
     padding-bottom: 25px;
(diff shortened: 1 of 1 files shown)


Reckless-Satoshi:
Any phone with issues to display the BookPage fully, probably also has issues with displaying the order details and trading prompts (buttons might lay outside).  Would it be possible to extend this fix (i.e. add vertical scroll on Safari) to what is displayed in the OrderPage? 

Need to test and merge. Thanks a lot!! :D

ozoono:
You are right. Order details page also has the same issues than the order book. Bottom buttons are not visible, and as there is no scroll , it is not possible to click them (i.e: when you want to confirm a payment)
This fix should solve the problem in both pages.

I can check it in the moment the changes are in Testnet

ozoono:
Styles updated to allow scrolling in Safari, into the page with the contract box, just like in the order book and order details page.

PR #131, opened 2022-05-08
Fix typos in spanish translation file

ozoono:
Fix some small typos in spanish translation file, and change a few words so they can be better understood


Reckless-Satoshi:
Very well spotted. Thank you!

frontend/src/locales/es.json
```diff
@@ -370,7 +370,7 @@
...
     "Is RoboSats private?": "¿RoboSats es privado?",
-    "RoboSats will never ask you for your name, country or ID. RoboSats does not custody your funds and does not care who you are. RoboSats does not collect or custody any personal data. For best anonymity use Tor Browser and access the .onion hidden service.": "RoboSats no custodia tus fondos y no recolecta o custodia ningún dato personal, pues no le importa quien eres. RoboSats nunca te preguntará por tu nombre, país o número de documento. Para mejorar tu privacidad, usa el Navegador Tor y visita el sitio cebolla.",
+    "RoboSats will never ask you for your name, country or ID. RoboSats does not custody your funds and does not care who you are. RoboSats does not collect or custody any personal data. For best anonymity use Tor Browser and access the .onion hidden service.": "RoboSats no custodia tus fondos y no recolecta o custodia ningún dato personal, pues no le importa quien eres. RoboSats nunca te preguntará por tu nombre, país o número de documento. Para mejorar tu privacidad, usa el Navegador Tor y accede al enlace .onion.",
```
ibertario, line 373: `sitio cebolla` is the spanish Tor Project official translation of Onion services / .onion site

Source https://tb-manual.torproject.org/es/onion-services/

ibertario (commented):
Thank you for the review! I made some mistakes and now it's much better.

frontend/src/locales/es.json
ozoono: Yes, you are right. I just proposed this alternative because not everybody knows what "servicio cebolla" means, while  "enlace .onion" seems something easier to understand.

Reckless-Satoshi: I think it does not matter which one we pick:) 

Sticking to the Tor Project terminology is a plus. But I had never heard "sitio cebolla" before you proposed it.The Tor Project probably does not have any linguistic professional on their team either :D. 

Onion is also straight forward to understand in Spanish since URLs finish with ".onion". It seems to be the most used term by the Spanish speakers in the wild.

ibertario: I vote in this case for `sitio .onion`! I will contact also to Tor Project and see... Cheers!

PR #142, opened 2022-05-18
change docpaths

smargold476:
correct paths


Reckless-Satoshi:
Thanks for catching this one. Merging it as a temporary fix. 

The new documentation site will be launch soon, probably tomorrow (https://learn.robosats.com/)

The new URL will be https://learn.robosats.com/read/en for written guides and /watch/en for video tutorials.

PR #143, opened 2022-05-18
Add more details in frontend setup guide about css files

ozoono:
Extend instructions about how-to locally test css files.

Current setup guide is valid if you want to test main.js file, but doesn't work when trying to locally edit and test changes in CSS files. This happens because if (according to the guide) you run the local server in "robosats/frontend/static/frontend/", CSS files are not going to be available, so the path has to be "robosats/frontend/static/"


Reckless-Satoshi:
Indeed it was only thought for place main.js . Much better now, thanks!

PR #152, opened 2022-05-29
translation polishing

Tonytherussian:


Reckless-Satoshi:
Hey @Tonytherussian I am unsure whether all your commits are actually already in the platform. I believe so. But can you confirm? Just check one of the last words/sentences you have editted.

If some of the commits are not in the platform, it would be very helpful to fork again from the current state of ru.json on the main branch and add the latest changes. The ru.json in this PR has json syntax errors and is in conflict with the one in main.

Thanks a lot for this super contrib!

Tonytherussian:
Hey!)
It's still confusing how github works for me. Somehow I've created many branches and was committing updates to different ones  :|
I see that there are some things missing, so will get to it now.

PR #156, opened 2022-06-01
Add currency aed, which Yadio now supports

sgornick:
For testing, I did nothing more than a sanity check on the following to ensure the rate was current:

https://api.yadio.io/json/AED


Reckless-Satoshi:
Thank you! Merging

Issue #128, opened 2022-05-08
Buyer payouts on-chain with on-the-fly reverse submarine swaps.


Reckless-Satoshi:
**Observation:**
New users on lightning seem to be confused about how they can move the Sats to an on-chain address or cold wallet. Usually, after buying sats in LN using RoboSats, they resort to some thrid party service (CoinOS, FxedFloat, Muun...) that charges a relatively large fee.

The swap service, however, might be easy to implement in RoboSats. A user could select to receive the Sats on-chain for an extra fee, and RoboSats will on-the-fly swap the LN Sats received from the seller. The user experience will greatly improve from the seamless swap, while RoboSats service becomes more vertically integrated and can generate revenue from the swaps improving the service sustainability.

**Feature requested:**
Add the option to submit an on-chain address instead of an invoice for buyer payouts.

**Some challenges** 
Will have to be faced regarding liquidity management. When RoboSats works 100% in Lightning, there is no need to close/open channels often. However, if users start to withdraw liquidity on-chain, the platform node will become a net sink. That will increase the cost of operating the node. There has to be a clear incentive for users to stay in lightning. therefore this feature will at first have a larger fee (~1-2%) to de incentivize its use. 

This will be a large feature that will need active maintainance. Many checks will have to be implemented (is this a valid on-chain address? Is there enough on-chain balance to allow this user the new option? Is this order large enough to be offered the option for seamless swap?) Also a new model "OnChainPayment" will be needed and a full fork of the Send Payment to Buyer pipeline both in backend and frontend.

Reckless-Satoshi:
Implemented and live
https://github.com/Reckless-Satoshi/robosats/pull/160

PR #160, opened 2022-06-05
Buyer payouts as on-the-fly reverse submarine swaps

Reckless-Satoshi:
This feature will allow buyers to submit an onchain address instead of an invoice. For a fee, they will be able to directly receive their Sats on their cold storage.

**Background and motivation in https://github.com/Reckless-Satoshi/robosats/issues/128#issue-1228870506**

TODO:
- [x] Implement LNnode functions to query channel, wallet and total balance.
- [x] DB model for "Liquidity" under the Control app
- [x] Implement LNnode functions to validate an onchain address and suggest fee in sats/vbyte
- [x] Implement utils for submarine swap cost: linearly proportional to available on-chain liquidity.
- [x] New API model for "OnchainPayment"
- [x] New Order props: is_swap (boolean), payout_onchain (foreign key OnchainPayment). 
- [x] Logic remake for on-chain address submission (instead of LNinvoice) and payment (instead of routing)
- [x] Add frontend component. Submit a payout invoice as a new tab: "submit onchain address". Only available if enough liquidity is onchain. Must inform user of % fee cost of the swap. Must suggest user a miner fee and ask for input (calculated in frontend assuming ~120 vbytes tx size). Ultimately shows user 2 amounts: exactly how many sats they receive and how many sats it costs.
- [x] Add frontend component. Transaction sent! show TXID. On successful trade window.


Reckless-Satoshi:
Ready and live in the testnet platform.

PR #172, opened 2022-06-25
Basque (eu) language added

4pt1x:


Reckless-Satoshi:
Awesome contribution. Looking forward to RoboSats in Basque :D

PR #173, opened 2022-06-25
Replace weighted average to weighted median.

satsbaba:
This pull request changes the weighted mean to the weighted median for the calculation of the premium displayed on the robosats.


Reckless-Satoshi:
Merged, thanks for this contribution!!

PR #178, opened 2022-06-28
Adding Portuguese translation

LindomarRodrigues:
I added the file frontend/src/locales/pt.json with the translation to Portuguese, I tried to be as neutral as possible, so possibly it will suit pt-PT and pt-BR.


Reckless-Satoshi:
Great contribution, looking forward to RoboSats in Portuguese!

PR #182, opened 2022-07-05
Basque lang few spelling errors corrected

OSFr0g:
Just a few small spelling errors solved, everything else looks great!


Reckless-Satoshi:
Thank you!

PR #185, opened 2022-07-17
Add Swedish translation

angeria:
## What
It's now possible to change the language of the website to Swedish. The translation is 100% complete.

## Why
To make RoboSats accessible to more people and improve the user experience.

### How to test it
You should be able to choose the Swedish flag from the language dropdown in the bottom-right corner, whereupon the language of the UI should immediately be swapped to Swedish.


Reckless-Satoshi:
Amazing, thank you very much!

Was just working on loading languages.json from http in order to decrease the app size, so I inadvertently creating a merge conflict with your work. Will fix it next!

PR #201, opened 2022-07-31
Compute percentile over orders of the same type

merchero:


Reckless-Satoshi:
Thanks @merchero better matches user expectation of what the premium rank is.

PR #209, opened 2022-08-09
Add grpc generation script and update setup.md to run it using docker container

### LINKED ISSUE #208: Improve docker-compose.yml / setup.md
Despite the single docker-compose file, it is still not easy to get started and there were quite a few pitfalls, at least for me

- `docker-compose build` copies the whole context and executes setup commands there, so they need to be run again locally (because the local files get mounted into the containers). This is mentioned in setup.md, but is not really convenient. Maybe this could be executed when a container is started (and not during the build). Or run a script using the backend image before running `docker-compose up`, no python setup would be needed on the host.
- `docker exec -it django-dev python3 manage.py makemigrations` only generated a migration for `api` (which failed to migrate because of missing tables from `control`), I needed to run `docker exec -it django-dev python3 manage.py makemigrations [control|chat]` separately to create all migrations. I'm not familiar with django so I don't know what went wrong. Also for me it wasn't clear, that the user created with `createsuperuser` must be in the `ESCROW_USERNAME` .env variable
- setup.md says _RoboSats development site should be accessible on localhost:8000_, but by default localhost is not a whitelisted host; one could use 127.0.0.1 instead so no configuration is necessary
- The path `/mnt/development` is quite specific and it may not be possible to use depending on the system configuration. Imo a generic location would be better (either a docker volume, or at least a directory within the repo; one could create a symlink to wherever large data should be stored)

These are all rather small problems, but they take time to solve and make getting started more complicated

ShatteredBunny:
Make it more convenient to setup Robosats locally. Fixes #208

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -46,6 +46,8 @@ Make "{robosats-site}/static/css/index.css" redirect to "127.0.0.1:8080/css/inde
 Spinning up docker for the first time
 ```
 docker-compose build --no-cache
+# Install LND python dependencies into local repository
+docker run --mount type=bind,src=$(pwd),dst=/usr/src/robosats backend sh generate_grpc.sh
 docker-compose up -d
 docker exec -it django-dev python3 manage.py makemigrations
 docker exec -it django-dev python3 manage.py migrate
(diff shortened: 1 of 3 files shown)


setup.md
```diff
@@ -46,6 +46,8 @@ Make "{robosats-site}/static/css/index.css" redirect to "127.0.0.1:8080/css/inde
 Spinning up docker for the first time
 ```
 docker-compose build --no-cache
+# Install LND python dependencies into local repository
+docker run --mount type=bind,src=$(pwd),dst=/usr/src/robosats backend sh generate_grpc.sh
```
Reckless-Satoshi, line 50: Great!

Reckless-Satoshi (approved):
It definitely cleans up the grpc install sequence as well as clear a lot of confusion for someone who runs the stack first time. Very valuable contribution.

PR #210, opened 2022-08-10
Stealth invoices

### LINKED ISSUE #168: Add 'stealth invoices': an On/Off toggle for lightning invoice description
**the privacy issue**
When paying lightning invoices for the bond or the maker trade amount, the super descriptive description in the lightning invoice is handy, BUT there are a lot of use cases where users doxing itself severely.

As an example, if an user pays the bond or trade amount from a custodian exchange (Kraken for example), Kraken is getting undeniable facts about the withdrawal. This is also true for all other custodial lightning wallets like Wallet of Satoshi, ...

This is probably true for invoices as a taker, but I didn't research into this side of the trade concerning this feature request so far.

**proposed solution**
Make adding a description to lightning invoices toggable for each user and default to NO description.

**alternatives**
I can't think of any alternatives to maintain privacy when interacting with robosats from custodial wallets.

ShatteredBunny:
Fixes #168 

The profile has now a flag whether stealth invoices should be used. The default is off.

When an order is created, a UUID is generated as a reference. This reference should never be public and only visible for the participants of the trade. If stealth invoices are enabled, only the reference is included into the invoice description.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -297,10 +297,7 @@ class Order(models.Model):
         NESINV = 4, "Neither escrow locked or invoice submitted"
 
     # order info
-    reference = models.CharField(max_length=36,
-                                 null=True,
-                                 default="",
-                                 blank=True)
+    reference = models.UUIDField(default = uuid.uuid4, editable = False)
     status = models.PositiveSmallIntegerField(choices=Status.choices,
                                               null=False,
                                               default=Status.WFB)
--- a/api/models.py
+++ b/api/models.py
@@ -297,6 +297,10 @@ class Order(models.Model):
         NESINV = 4, "Neither escrow locked or invoice submitted"
 
     # order info
+    reference = models.CharField(max_length=36,
+                                 null=True,
+                                 default="",
+                                 blank=True)
     status = models.PositiveSmallIntegerField(choices=Status.choices,
                                               null=False,
                                               default=Status.WFB)
--- a/api/models.py
+++ b/api/models.py
@@ -652,6 +652,10 @@ class Profile(models.Model):
                                                   default=None,
                                                   blank=True)
 
+    # Stealth invoices
+    wants_stealth = models.BooleanField(default=False,
+                                        null=False)
+
     @receiver(post_save, sender=User)
     def create_user_profile(sender, instance, created, **kwargs):
         if created:
--- a/api/views.py
+++ b/api/views.py
@@ -30,6 +30,7 @@ from datetime import timedelta, datetime
 from django.utils import timezone
 from django.conf import settings
 from decouple import config
+from uuid import uuid4
 
 EXP_MAKER_BOND_INVOICE = int(config("EXP_MAKER_BOND_INVOICE"))
 RETRY_TIME = int(config("RETRY_TIME"))
@@ -141,6 +142,7 @@ class MakerView(CreateAPIView):
             escrow_duration=escrow_duration,
             bond_size=bond_size,
             bondless_taker=bondless_taker,
+            reference=uuid4()
         )
 
         order.last_satoshis = order.t0_satoshis = Logics.satoshis_now(order)
--- a/api/views.py
+++ b/api/views.py
@@ -30,7 +30,6 @@ from datetime import timedelta, datetime
 from django.utils import timezone
 from django.conf import settings
 from decouple import config
-from uuid import uuid4
 
 EXP_MAKER_BOND_INVOICE = int(config("EXP_MAKER_BOND_INVOICE"))
 RETRY_TIME = int(config("RETRY_TIME"))
@@ -142,7 +141,6 @@ class MakerView(CreateAPIView):
             escrow_duration=escrow_duration,
             bond_size=bond_size,
             bondless_taker=bondless_taker,
-            reference=uuid4()
         )
 
         order.last_satoshis = order.t0_satoshis = Logics.satoshis_now(order)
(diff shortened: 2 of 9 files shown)


api/models.py
```diff
@@ -297,6 +297,10 @@ class ExpiryReasons(models.IntegerChoices):
         NESINV = 4, "Neither escrow locked or invoice submitted"
 
     # order info
+    reference = models.CharField(max_length=36,
```
Reckless-Satoshi, line 300: Might be cleaner if we use a `UUIDField` instead of a `CharField`. It will be filled with a UUID as soon as the object is created  (more info on UUIDFields https://www.geeksforgeeks.org/uuidfield-django-models/ )
```python
import uuid
  
reference = models.UUIDField(
         default = uuid.uuid4,
         editable = False)
```

Would save a few lines also in views.py

api/views.py
```diff
@@ -30,6 +30,7 @@
 from django.utils import timezone
 from django.conf import settings
 from decouple import config
+from uuid import uuid4
```
Reckless-Satoshi, line 33: Not needed if we use a UUIDField

api/views.py
```diff
@@ -141,6 +142,7 @@ def post(self, request):
             escrow_duration=escrow_duration,
             bond_size=bond_size,
             bondless_taker=bondless_taker,
+            reference=uuid4()
```
Reckless-Satoshi, line 145: Same as above

api/views.py
```diff
@@ -740,10 +742,13 @@ def post(self, request, format=None):
                 user.profile.is_referred = True
                 user.profile.referred_by = queryset[0]
 
+            user.profile.wants_stealth = False
```
Reckless-Satoshi, line 745: Maybe good idea to leave here. But I believe it is not needed since `user.profile.wants_stealth` is already `False` by default

frontend/src/components/Dialogs/Profile.tsx
```diff
@@ -206,6 +210,22 @@ const ProfileDialog = ({
 
           <Divider/>
 
+          <Grid container>
+            <Grid item>
+              <FormControlLabel
+                labelPlacement="start"
+                label={`${t("Use stealth invoices")}`}
```
Reckless-Satoshi, line 217: The translation key `"use stealth invoices":"{same but in another language}"` is now available. Would be ideal to add a new key to `en.json` so future translations include it. It can be added at the end of "Bottom Bar", around here:
https://github.com/Reckless-Satoshi/robosats/blob/efd89756e41820cd0a1cc26105d23a7300094f84/frontend/static/locales/en.json#L229-L230

api/views.py
```diff
@@ -740,10 +742,13 @@ def post(self, request, format=None):
...
+
             user.profile.save()
 
             context["public_key"] = user.profile.public_key
             context["encrypted_private_key"] = user.profile.encrypted_private_key
+            context["wants_stealth"] = False
```
Reckless-Satoshi, line 751: Looks good, but if we remove the line above it would make sense to change this one to
```python

context["wants_stealth"] = user.profile.wants_stealth
```

Reckless-Satoshi (changes requested):
This is a superb work! tested and works flawlessly.

I left a few minor comments in case you want to polish the few things I've detected. Most are really not needed, just in case you feel they add to your work.

I am very excited to see you got around so quickly to run the stack and contribute both to back and frontend. Welcome to robodevs and thank you!! 

Please submit an invoice for 120K Sats from a proxy wallet (not your own node_id) and a long expiration time (+2 days).

ShatteredBunny:
Thanks for going through it, I didn't knew about uuid fields and missed the localization file
I also added a tooltip explaining what stealth invoices do, may not be obvious

This invoice is only valid one day, I don't know if there is a custodial wallet that lets you set the time... let me know if it was too short
`lnbc1200u1p30dz79pp5kh93tv9uq7m5v346kkxtqhc8fqmh2ms7sh6wtn4yndrzayu6pj5qdqu2askcmr9wssx7e3q2dshgmmndp5scqzpgxqyz5vqsp5mwl30w504agyktpkft2ujy3l0m50d6hnwqy8vue4984z0zd322jq9qyyssqsjgxrw0dpr06rp7ev0d6xdlzke060dcyeslcqd083fvz9nz79pf9mejazfnmfnyjk4jthx304k0p84nszjcjq50edt3qhqg273l4wegp4w3g3x`

Reckless-Satoshi:
Looks great and the Tooltip will indeed be very handy !

> `lnbc1200u1p30dz79pp5kh93tv9uq7m5v346kkxtqhc8fqmh2ms7sh6wtn4yndrzayu6pj5qdqu2askcmr9wssx7e3q2dshgmmndp5scqzpgxqyz5vqsp5mwl30w504agyktpkft2ujy3l0m50d6hnwqy8vue4984z0zd322jq9qyyssqsjgxrw0dpr06rp7ev0d6xdlzke060dcyeslcqd083fvz9nz79pf9mejazfnmfnyjk4jthx304k0p84nszjcjq50edt3qhqg273l4wegp4w3g3x`

1ff79dbae08332ecd811f750b4f00dd007145bbf7ad80274da12981ee57a35b7

Merging!

Reckless-Satoshi:
Looks great and the Tooltip will indeed be very handy !

> `lnbc1200u1p30dz79pp5kh93tv9uq7m5v346kkxtqhc8fqmh2ms7sh6wtn4yndrzayu6pj5qdqu2askcmr9wssx7e3q2dshgmmndp5scqzpgxqyz5vqsp5mwl30w504agyktpkft2ujy3l0m50d6hnwqy8vue4984z0zd322jq9qyyssqsjgxrw0dpr06rp7ev0d6xdlzke060dcyeslcqd083fvz9nz79pf9mejazfnmfnyjk4jthx304k0p84nszjcjq50edt3qhqg273l4wegp4w3g3x`

1ff79dbae08332ecd811f750b4f00dd007145bbf7ad80274da12981ee57a35b7

Merging!

PR #212, opened 2022-08-12
Make robosats setup more easy

ShatteredBunny:
I just fixed the other problems mentioned in #208 
Setup should be much easier now
- Add apps `api`, `control` and `chat` to `makemigrations`
- Remove absolute path `/mnt/development` and use `./node` instead; you may want to create a symlink instead (`ln -s /mnt/development node`)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,3 +10,4 @@ setup.md
 docker-compose.yml
 .github
 .git
+node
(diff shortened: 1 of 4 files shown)


Reckless-Satoshi:
Looks good! A very needed improvement of the development environment setup

PR #213, opened 2022-08-15
keeping russian up to date

Tonytherussian:


Reckless-Satoshi:
Thanks for adding all of the new keys to the Russian translation :rocket:

Issue #134, opened 2022-05-13
Add book depth chart


Reckless-Satoshi:
Create a book depth chart as a react component that fetches data from [`/api/book/?currency=0&type=2`](http://unsafe.robosats.com/api/book/?currency=0&type=2&format=json).`
It could use plotly/chart.js/canvas.js . 

A reference of what a usual book depth plot looks like:
![1*NjNZwgGH-jisDvbk7GvrOw-3617449492](https://user-images.githubusercontent.com/90936742/168276905-caff58fe-fe94-4091-b0eb-dc60d7f12455.png)

It could be subsettable for currencies (only show liquidity for a single currency) or aggregate all currency offers (e.g. convert every fiat to USD)or a better overview of the whole market. On the the X axis it could show price denominated in the fiat currency, or premium (%) over external market prices.

This task is currently rewarded with **⚡ 300,000 Sats** . Reply to this comment to be assigned.

PeterMcBTC:
apparently the code is described in the following article https://towardsdatascience.com/learn-what-a-depth-chart-is-and-how-to-create-it-in-python-323d065e6f86
I will have a look at it

Reckless-Satoshi:
@PeterMcBTC Awesome! The tutorial seems to be python based. It would certainly be more useful if the plot can be generated in the frontend in javascript with data fetched from unsafe.robosats.com /api/book/?currency=0&type=2.

KoalaSat:
Is there any preference on the charts library? 

I have good experiences with https://nivo.rocks, it add a really good abstraction layer to d3 yet it still allows you to go as custom as you want.

Specifically I think a Line will do a good fit with this case:
- https://nivo.rocks/line/
- https://nivo.rocks/storybook/?path=/story/line--stacked-lines
- https://nivo.rocks/storybook/?path=/story/line--real-time-chart

KoalaSat:
Also, I believe this is going to kill the back-end if we don't implement WebSockets, I did a quick research and looks like it's possible to run it over TorNet https://gist.github.com/TooTallNate/6253581

Reckless-Satoshi:
> Is there any preference on the charts library?

Something pretty and easy :) This is a very creative task, imposing a library wouldn't make sense.

>    Also, I believe this is going to kill the back-end if we don't implement WebSockets

[...]

PR #219, opened 2022-08-25
Add book depth chart

KoalaSat:
Fixes https://github.com/Reckless-Satoshi/robosats/issues/134

Screenshot 2022-08-26 at 09 38 11
image

## Max Zoom
Lines disappears after over passing Y Axis

image

## Cross-hair
image

## Mobile
Screenshot 2022-08-26 at 09 38 29

## Y Axis
`````
const calculateBtc = (order: any): number => {
  const amount = parseInt(order.amount || order.max_amount)
  return amount / order.price
}
`````

## X Axis
``````
order.premium
``````

### CODE REVIEW

File: frontend/src/components/Charts/DepthChart/index.tsx [hunk 1]
```diff
@@ -0,0 +1,172 @@
+import React, { useEffect, useState } from "react"
```
KoalaSat, line 1: Is a good practice in React to always create `index.ts|tsx `files, let me know what do you think

frontend/src/components/Charts/nivoScheme.ts
```diff
@@ -0,0 +1,29 @@
+import { Theme } from "@nivo/core"
+
```
KoalaSat, line 2: This is intended to be the default shared theme for future Nivo charts

frontend/src/components/Charts/nivoScheme.ts
```diff
@@ -0,0 +1,29 @@
...
+export const nivoScheme: Theme = {
+  axis: {
+    ticks: {
+      text: {
+        fill: "rgb(255, 255, 255)"
```
Reckless-Satoshi, line 7: Would be great if these variables can be dynamically set to match the colors defined in the MUI theme in `App.js` . It is provided as context to all children. https://github.com/Reckless-Satoshi/robosats/blob/8eda1dd20fec576726cb1578e263bfb43472be48/frontend/src/components/App.js#L26-L42
This would solve the conflict we have now with the light/dark themes as well as make it future-proof for any possible theme change.

Take a look at MUI Theming and the use of the hook `useTheme()` https://mui.com/material-ui/customization/theming/

frontend/src/components/BookPage.js
```diff
@@ -460,49 +525,31 @@ class BookPage extends Component {
...
-            }
-            <Button color="secondary" variant="contained" to="/" component={Link}>
-                {t("Back")}
-            </Button>
+            <ButtonGroup variant="contained" aria-label="outlined primary button group">
+              { !this.props.bookNotFound ?
+                <>
+                  <Button variant="contained" color='primary' to='/make/' component={Link}>{t("Make Order")}</Button>
+                  <Button color='inherit' style={{color: '#111111'}} onClick={this.handleClickView}>
+                    { this.state.view == 'depth' ? t("List") : t("Depth") }
```
Reckless-Satoshi, line 544: What do you think of using "Chart" instead of "Depth"? 
We could also add a small icon on this button so it is highlighted. For example:
```javascript
// Chart Button
import BarChartIcon from '@mui/icons-material/BarChart';
// List Button
import FormatListBulletedIcon from '@mui/icons-material/FormatListBulleted';
```

Reckless-Satoshi (commented):
This is perfect and really fun to use! Congratulations I am sure Robots will enjoy this feature. I certainly do :smile:

- The chart theme is not sensitive to a switch between dark and light theme on the RoboSats app. The axes and numbers can only be seen with the dark theme (therefore not visible by default in Tor Browser). The lines themselves would also look prettier on the light theme if they were darker.
- I am unsure on how the (-) and (+) buttons are supposed to work. These are zoom-in and zoom-out buttons, but at the moment they do not seem to be specially useful, do you envision them being useful at some point? 
- Given that the last day premium is located on the center as if it was the chart title, one instantly think that it is also the center of the Depth Chart, and that - and + will move you up and down in the X axis.
- I see finally you have left the X axes only in premium units. What happened to the X axes priced in Fiat units? :D Both are potentially very useful and informative. In fact, the fiat unit chosen for display could be the one in `this.props.currency`, to not leave behind those users who do not price life in USD.

All in all I think the chart might not need to display the last day average premium (or at least, not so prominently). What do you think?

Hypothetical cool things that could be done:

- On desktop, hovering the mouse shows a crosshair and the points can be "clicked". I am not sure if this is even possible, since the chart is aggregating the height of many orders... Is there a way we can overlay a small Tooltip with a little info about the order that matches that premium and has the highest volume on mouse hovering (E.g. RobotName, FiatAmount, Currency and Payment Method Icons)? That would be really top!!

KoalaSat:
> I am unsure on how the (-) and (+) buttons are supposed to work. These are zoom-in and zoom-out buttons, but at the moment they do not seem to be specially useful, do you envision them being useful at some point?

I just thought + and - represent the range, but you are right I checked other exchanges and it's clearly a zoom

> Given that the last day premium is located on the center as if it was the chart title, one instantly think that it is also the center of the Depth Chart, and that - and + will move you up and down in the X axis.
> All in all I think the chart might not need to display the last day average premium (or at least, not so prominently). What do you think?

The point is, that value is the one that determines de exact middle X point on the chart, Y can hide it but I see it useful to easily understand where the chart is located.

> On desktop, hovering the mouse shows a crosshair and the points can be "clicked". I am not sure if this is even possible, since the chart is aggregating the height of many orders... Is there a way we can overlay a small Tooltip with a little info about the order that matches that premium and has the highest volume on mouse hovering (E.g. RobotName, FiatAmount, Currency and Payment Method Icons)? That would be really top!!

I think with the magic of Nivo this should be easy peasy

KoalaSat:
> I see finally you have left the X axis only in premium units. What happened to the X axes priced in Fiat units? :D Both are potentially very useful and informative. In fact, the fiat unit chosen for display could be the one in this.props.currency, to not leave behind those users who do not price life in USD.

I understood the idea was just show, the premium, no problem I have that piece of code I can recover it

Reckless-Satoshi:
> The point is, that value is the one that determines de exact middle X point on the chart, Y can hide it but I see it useful to easily understand where the chart is located.

Ah, I see! Maybe we can instead make a mark for the "last 24h premium" as a dotted vertical line in the chart and label it? It would be clearer in my opinion.

KoalaSat:
> > The point is, that value is the one that determines de exact middle X point on the chart, Y can hide it but I see it useful to easily understand where the chart is located.
> 
> Ah, I see! Maybe we can instead mark the "last 24h premium" as a dotted vertical line in the chart and label it inside the chart? It would be clearer in my opinion. Centered texts just above a chart usually are the title.

That was my original idea but there is no way to make it appear 🥲 the documentation clearly allows it and it had an example but after trying over and over I found to way, I'll give another try with a workaround I don't like but maybe it'll make it work

KoalaSat:
New Version looks cooler

## X Axis selector
image
image

- When the selection is `Price`, the amount is calculated over the base currency selected (USD if ANY)
- The Price middle point id the median of all the converted amounts

## Tooltip
image

As suggested, I have included a too-tip per point, the criteria to show one point or another is based on a boronoi map generated by Nivo. Click on the chart when a tool-tip is being displayed pushes the user to the order page.

**IMPORTANT** Because I wanted to reuse the avatar code, I ended up refactoring a little bit the columns arrays on the book orders list, I hope that's fine, there was a couple of unnecessary loops, another unnecessary serialization and a lot of duplicated/unused code. Right now I think it looks cleaner :)

## Light/Dark Mode

image

A function returns one theme or another base on Mui's theme, I decided to move the entire object in case in the future we want to reuse colors.

## Vertical middle line

I didn't manage to make it work 😅

frontend/src/components/BookPage.js
```diff
@@ -135,77 +129,65 @@ class BookPage extends Component {
...
-              type: order.type ? t("Seller"): t("Buyer"),
-              amount: order.amount,
-              has_range: order.has_range,
-              min_amount: order.min_amount,
-              max_amount: order.max_amount,
-              currency: this.getCurrencyCode(order.currency),
-              payment_method: order.payment_method,
-              price: order.price,
-              premium: order.premium,
-            })
```
KoalaSat, line 154: I consider this serialization unnecessary, we already have an `book` object with all the information, I see no point on just changing names and confuse future developers

frontend/src/components/BookPage.js
```diff
@@ -237,80 +219,60 @@ class BookPage extends Component {
...
           { field: 'premium', headerName: t("Premium"), type: 'number', width: 85,
             renderCell: (params) => {return (
-              <Tooltip placement="left" enterTouchDelay={0} title={pn(params.row.price) + " " +params.row.currency+ "/BTC" }>
+              <Tooltip placement="left" enterTouchDelay={0} title={pn(params.row.price) + " " + params.row.currency + "/BTC" }>
               <div style={{ cursor: "pointer" }}>{parseFloat(parseFloat(params.row.premium).toFixed(4))+"%" }</div>
               </Tooltip>
-            )} },
-          ]}
+            )} 
+          }]}
```
KoalaSat, line 275: In another iteration I see this array reusable for both tables mobile and desktop

frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+    ) : <></>
+  }
+
+  const handleOnClick: PointMouseHandler = (point: Point) => {
+    history.push('/order/' + point.data?.order?.id);
+  }
+  
+  return bookLoading || !center || enrichedOrders.length < 1 ? (
+    <div style={{display: "flex", justifyContent: "center", paddingTop: 200, height: 420 }}>
+      <CircularProgress />
```
Reckless-Satoshi, line 227: The chart feels sluggish to load if you switch a few times between List/Chart repeatedly. I believe this is because every single time the chart is loaded, there is a call to "/limits/" that has to be waited.

Do you think we could save `/limits/` into the app state and from that point on only update that variable? (same as we do with the book orders)

frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+                <Grid item xs={12}>
+                  {amountToString(order.amount, order.has_range, order.min_amount, order.max_amount)}
+                  {' '}
+                  {currencyDict[order.currency]}
+                </Grid>
+                <Grid item xs={12}>
+                  <PaymentText 
+                    othersText={t("Others")} 
+                    verbose={true} 
+                    size={12}
```
Reckless-Satoshi, line 209: I would maybe use size 20 or 22 for better visibility.  Any cons to this bigger size?

frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+          }}
+          axisBottom={{
+            tickSize: 5,
+            format: formatAxisX
+          }}
+          margin={{ left: 65, right: 60, bottom: 25, top: 10 }}
+          xFormat={(value) => Number(value).toFixed(0)}
+          lineWidth={3}
+          theme={getNivoScheme(theme)}
+          colors={['rgb(136, 252, 102)', 'rgb(255, 108, 57)']}
```
Reckless-Satoshi, line 308: While orange and green is widely used for depth charts, we could give it a touch by sticking to the RoboSats color scheme. Throughout the UI we are loosely sticking to using purple for "things you click if you want to sell" and blue for "things you click on if you want to buy".

What do you think of this?
``` javascript
          colors={[theme.palette.secondary.main,theme.palette.primary.main]}
```
These are the color references for the default MUI palette https://mui.com/material-ui/customization/palette/
When using these theme variables, the line colors will be brighter when the dark mode is enabled.

Reckless-Satoshi (changes requested):
This chart is soo much fun to use and look at! To me is already looking ready for merge. I requested a few tiny changes.

A couple of things we also might want to check:
- Switching the currency on the drop down menu does change the currency code shown on the chart, but does not change the tick labels of the X axis (They remain in USD)
- In mobile, when using Price in the X axis, the ticks of the X label become unreadable as they are too close to each other. Maybe the number of ticks has to be decimated.
- When hovering the mouse, for a fraction of a second, the tooltips show below the pointer, then on top. No issue whatsoever, yet when you are looking at tooltips on the bottom of the chart, the bigger size makes the vertical and horizontal scroll bars show. Just a tiny detail that might be too hard to fix to bother though...
chart
(I was able to take this screenshot, but it lasts for a fraction of a second only)

All in all congrats! This is a feature robots will love :rocket:

KoalaSat:
> When hovering the mouse, for a fraction of a second, the tooltips show below the pointer, then on top. No issue whatsoever, yet when you are looking at tooltips on the bottom of the chart, the bigger size makes the vertical and horizontal scroll bars show. Just a tiny detail that might be too hard to fix to bother though...

looks like something on Nivo, it only happens for the first tool tip that loads once the mouse hovers the charts, looks like the default position, which quickly gets the first configuration and changes, I don't see any way to change it from outside the library.

frontend/src/components/Charts/DepthChart/index.tsx
KoalaSat: Good catch! I change it and now it only loads once :)

KoalaSat:
Good news, I managed to generate the middle line 😄 

image

I think I already covered all your comments @Reckless-Satoshi 

> In mobile, when using Price in the X axis, the ticks of the X label become unreadable as they are too close to each other. Maybe the number of ticks has to be decimated.

This is my proposal for when the chart is displayed on mobile
image

KoalaSat:
BTW this is my proposal to avoid overlapping on mobile view

image

Reckless-Satoshi:
It's looking great!

> BTW this is my proposal to avoid overlapping on mobile view
> 
> image

It's looking great!

I will take a more deep read at the code to understand myself everything that has been introduced before merging. So it might take 1-2 days until this PR is merge.

Please, share a LN invoice for 300K with a long expiration time and from a proxy wallet.

KoalaSat:
> I will take a more deep read at the code to understand myself everything that has been introduced before merging. So it might take 1-2 days until this PR is merge.

Take your time, any clarification or modification request will be welcome

KoalaSat:
> Please, share a LN invoice for 300K with a long expiration time and from a proxy wallet.

lnbc3m1p3shvrlpp5wtptd958lwqwekgrdctyf5geuzwu6j380wwqcpppqj0dhu2h2xesdq5g9kxy7fqd9h8vmmfvdjscqzpgxqyz5vqsp55mhyf3g3vadlnkcpuuamanlmc2npglwx5kr2dcqjslmdeunar73s9qyyssq3l20e72s82y6nlvfx2h6lp4cwwf64frslk8x928yvsqwyr0y66dkt389xku0tr7z26t8n4qqgqwuamu7phcnjyx4e73xez2u4y255kqqppql8a

frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -0,0 +1,346 @@
...
+    } else if (lastDayPremium) {
+      setCenter(lastDayPremium)
+      setXRange(8)
+    }
+  }, [enrichedOrders, xType, lastDayPremium, currencyCode])
+
+  const calculateBtc = (order: Order): number => {
+    const amount = parseInt(order.amount) || order.max_amount
+    return amount / order.price
+  }
```
KoalaSat, line 87: This calculus might potentially be included on the back-end response

frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -0,0 +1,346 @@
...
+            </IconButton>
+          </Grid>
+        </Grid>
+      </Grid>
+      <Grid container style={{ height: 357, padding: 15 }}>
+        <ResponsiveLine 
+          data={series} 
+          enableArea={true}
+          useMesh={true}
+          animate={false}
```
KoalaSat, line 312: In general, the animation feature is cool, but on this case the X axis scale changes so drastically (from premium to price) that it really looks odd.

frontend/src/components/Charts/DepthChart/index.tsx
Reckless-Satoshi: It can be added to the API. It is still very fluid.

Reckless-Satoshi:
> > Please, share a LN invoice for 300K with a long expiration time and from a proxy wallet.
> 
> lnbc3m1p3shvrlpp5wtptd958lwqwekgrdctyf5geuzwu6j380wwqcpppqj0dhu2h2xesdq5g9kxy7fqd9h8vmmfvdjscqzpgxqyz5vqsp55mhyf3g3vadlnkcpuuamanlmc2npglwx5kr2dcqjslmdeunar73s9qyyssq3l20e72s82y6nlvfx2h6lp4cwwf64frslk8x928yvsqwyr0y66dkt389xku0tr7z26t8n4qqgqwuamu7phcnjyx4e73xez2u4y255kqqppql8a

3037303861373334363436653765376562316364343532383730623833633163

(Funny, Alby wallet preimage HEX is always made only with decimal numbers!)

KoalaSat:
> > > Please, share a LN invoice for 300K with a long expiration time and from a proxy wallet.
> > 
> > 
> > lnbc3m1p3shvrlpp5wtptd958lwqwekgrdctyf5geuzwu6j380wwqcpppqj0dhu2h2xesdq5g9kxy7fqd9h8vmmfvdjscqzpgxqyz5vqsp55mhyf3g3vadlnkcpuuamanlmc2npglwx5kr2dcqjslmdeunar73s9qyyssq3l20e72s82y6nlvfx2h6lp4cwwf64frslk8x928yvsqwyr0y66dkt389xku0tr7z26t8n4qqgqwuamu7phcnjyx4e73xez2u4y255kqqppql8a
> 
> 3037303861373334363436653765376562316364343532383730623833633163
> 
> (Funny, Alby wallet preimage HEX is always made only with decimal numbers!)

I was also wondering what were the odds the first time 😄

Reckless-Satoshi (approved):

frontend/src/components/Charts/DepthChart/index.tsx
KoalaSat: I was thinking on creating a separate PR after we merge this one so we don't overcharge it, what do you think?

Reckless-Satoshi: Yes, let's do that on a different PR. Take a look at Model order and the var `order.last_satoshis`, it keeps a snapshot of the size of the order in satoshis when it is created. Is this enough or do we need it to be *realtime* ?

KoalaSat: It has to be real time, consider everyone creates right their orders and this night Bitcoin pumps up 20%, the amount of BTC you'll be able to buy tomorrow morning is going to be 20% more than the real quantity.

KoalaSat: I found these 2:

https://github.com/Reckless-Satoshi/robosats/blob/eff58dc91d93fe98a6738e0e3898ed2a476d2ebc/api/models.py#L382
https://github.com/Reckless-Satoshi/robosats/blob/eff58dc91d93fe98a6738e0e3898ed2a476d2ebc/api/models.py#L390

but  the only logic for `t0_satoshis` is this:

https://github.com/Reckless-Satoshi/robosats/blob/eff58dc91d93fe98a6738e0e3898ed2a476d2ebc/api/views.py#L146

Looks like there was an original intention to keep the sats on creation and the sats right now, that's cool, but at some point it got lost and `last_satoshis` was used as `t0_satoshis`.

I like the original idea, I had a look and there are not too many places where `last_satoshis` is being used, do you believe it's a good idea to try a renaming?

KoalaSat: I think I'll move this conversation to an issue I can assign to myself so it doesn't get lost into this PR

frontend/src/utils/prettyNumbers.test.ts
```diff
@@ -24,3 +24,22 @@ describe("prettyNumbers", () => {
     });
   });
 })
+
+describe("amountToString", () => {
+  test("pn()", () => {
+    [
+      {input: null, output: undefined},
```
Reckless-Satoshi, line 31: @KoalaSat just a heads up. After merging this PR the frontend build GH action was failing because it did not pass this test. I amended this test in the main branch.

PR #220, opened 2022-08-26
Auto-scale UI: UserGen example

fmitjans:
Working example for the UserGen page. Let me know if solution is ok to do the same on the other pages

### DIFF
--- a/frontend/src/components/UserGenPage.js
+++ b/frontend/src/components/UserGenPage.js
@@ -182,29 +182,32 @@ class UserGenPage extends Component {
   render() {
     const { t, i18n} = this.props;
     const fontSize = this.props.theme.typography.fontSize;
+    const fontSizeFactor = fontSize / 14; // to scale sizes, default fontSize is 14
     return (
       <Grid container spacing={1}>
         <Grid item>
           <div className='clickTrough'/>
         </Grid>
-        <Grid item xs={12} align="center" sx={{width:370, height:260}}>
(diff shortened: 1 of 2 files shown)


frontend/src/components/UserGenPage.js
```diff
@@ -182,29 +182,32 @@ class UserGenPage extends Component {
   render() {
     const { t, i18n} = this.props;
     const fontSize = this.props.theme.typography.fontSize;
+    const fontSizeFactor = fontSize / 14; // to scale sizes, default fontSize is 14
```
Reckless-Satoshi, line 185: Clever :)

Reckless-Satoshi (commented):
Tested, works exactly as expected! :rocket: 

We won't be merging this PR. But this is exactly what has to be done in the other components to scale the UI. 

Be mindful a few sizes are harcoded in the CSS files. These have to be moved to an `style` or `sx` prop. For example, the height of the BottomBar: https://github.com/Reckless-Satoshi/robosats/blob/8eda1dd20fec576726cb1578e263bfb43472be48/frontend/static/css/index.css#L61-L66

To avoid clutter of the UI, for the final PR it would be very appreciated if we also can hide the zoom-in zoom-out buttons inside of a new Dialog "Settings". The access to this dialog could be placed exactly in location where the dark/light theme switch is now and use the icon `SettingsIcon` from `'@mui/icons-material/Settings';` . The dialog will contain the zoom-in, zoom-out, dark/light theme toggle button.

Reckless-Satoshi:
Looking great! Closing without merge as it is not production ready.

PR #222, opened 2022-08-27
Use linter and formatter

ShatteredBunny:
To enforce a consistent code style a linter and formatter should be used.

The linter was already configured, I mainly added scripts to run it. Additionally, I added prettier as a formatter.
Using `npm run format` (or `docker exec -it npm-dev npm run format`), the code could be formatted automatically.
I just grabbed some defaults, like semicolons at the end of lines, single quotation marks for strings and trailing commas.

Side mark: the package-lock.json was generated using an old node/npm version. I ran `npm install` using the docker container, but `--force` was needed because `@mui` does not officially support React 18 yet…


Reckless-Satoshi:
Hey @ShatteredBunny, thanks for taking care of this!

> Side mark: the package-lock.json was generated using an old node/npm version. I ran `npm install` using the docker container, but `--force` was needed because `@mui` does not officially support React 18 yet…

Might this have been caused by a breaking change introduced when switching the base image from `ubuntu:20.03` to `node:18-bullseye-slim` ? https://github.com/Reckless-Satoshi/robosats/commit/45fd82430ed501caca3f2d67f0d63b13fce6519c#diff-ea60ef29f6f537c1c83468c00b60de28f86e06edfe4a3d91274723c6f298fdb8L1-L3 I noticed the `package-lock.json` warning, did not pay much attention to it as everything else worked as usual.

ShatteredBunny:
> Might this have been caused by a breaking change introduced when switching the base image

Indirectly, the Ubuntu image used a quite old version of node and npm, I think `npm install` just gave a warning there that peer dependencies are incompatible.
The new node version (node image or local) requires `--force` to ignore the warning. The problem should be fixed in the next version of material-ui, and it seems to be working when ignoring the warning...

The typescript linter doesn't even run on the old node version, so I'd stick with the new version

Reckless-Satoshi (commented):
Can we only include in this PR the changes to `.eslinctrc.json` , `.prettierignore`,  `.prettierrc` `package.json` `package-lock.json` and `CONTRIBUTING.md` ?

That will make merging this PR easier. After that, I would run `npm run format` the first time and then commit. I deem unsafe to merge a PR that changes so many lines I cannot really spend the time to glance over.


Q: What do you think about adding a Github actions workflow that runs `npm run format` for every PR changing `/frontend/`? Maybe this does the trick https://github.com/marketplace/actions/prettier-action

ShatteredBunny:
> Can we only include in this PR the changes

Sure, good idea, will do it later

I don't know how github actions work, will formatter changes be added to every merge? Sounds good to me, I thought about letting the linter complain about wrong formatting and letting checks fail (in the future, there are still a few hundred linter issues), but that would of course be much more lets say developer friendly

Reckless-Satoshi:
Hey @ShatteredBunny I merged this PR. However, I have now reverted master to the previous commit.

This PR introduced the breaking change of having to `npm install --force`, for local testing that might be okay. However, when it comes to the CI/CD pipeline to build the next production robosats frontend --force is also needed. This is probably not great, since we might keep introducing unintendedly more incompatible packages with breaking changes down the line and we will not notice.

Is there a way we can introduce the linter / formatter on the current codebase without a need to use `--force` on the install? We definitely should use the linter.

I also tested the linter GH actions to little success so far :)

ShatteredBunny:
I see, that is not good. But the repository is kind of in a broken state...
I realized a weird behaviour of npm, running npm install while the old package-lock.json is there works, running it a second time fails. So the current 'fix' is not committing the package-lock.json. Probably this way conflicting packages will also stay undiscovered

Options I see:
 - Downgrade React to version 17 to fix the dependency error
 - Using --force until material updated to React 18 and be careful when merging
 - Exclude package-lock.json, probably same effect as --force, but as a weird hack

Reckless-Satoshi:
> Options I see:
> 
>     * Downgrade React to version 17 to fix the dependency error
> 
>     * Using --force until material updated to React 18 and be careful when merging
> 
>     * Exclude package-lock.json, probably same effect as --force, but as a weird hack

I researched a bit and I have found one more option that is long term preferred.

The dependency error is due to MUI v4 libraries, yet we mostly are using MUI v5. I doubt there will be an update of the old MUI v4 to make it react 18.0.0 compatible, so we might prefer to just drop these libraries. We only need them since they are required by `material-ui-image` ( and we use this one for the fade-in loading animation of images with a spinner for Robot Avatars). I am sure we can find an alternative that is better maintained :smile: 

So we drop:
```json
    "@material-ui/core": "^4.12.3",
    "@material-ui/icons": "^4.11.2",
    "material-ui-image": "^3.3.2",
    "react-qr-reader": "^2.2.1", //has never worked and also relies on <18.0.0)
```
And then we update react and add a missing dependency
```json
    "react": "^18.2.0",
    "@date-io/date-fns": "^2.15.0", // was missing on package.lock
```

Tested and this should install without conflicts. In addition dropping those four dependencies has a dramatic effect on the size of the bundled `main.js`! Should improve loading time by ~30%. It's a final solution to the dependencies problem, what do you think?

I can look into finding the `material-ui-image` replacement later this week. Or we can make it a small rewarded task (80K Sats) if you want to put hands into it. The task looks like this:

- [ ] Replace use of `Image` from `material-ui-image` in `src/components/UserGenPage.js`, `src/components/BookPage.js` and `src/components/Dialogs/Info.tsx` for an alternative component (a simple fade-in image loader optionally with a spinner)
- [ ] Remove the use of `QrReader from 'react-qr-reader'` and the QR scan button from `src/components/TradeBox.js`
- [ ] Remove and add the dependencies above from `package.json` and generate a new `package-lock.json`

ShatteredBunny:
Okay, I haven't looked at it in that much detail, I'd agree that getting rid of the unsupported dependencies should be the goal and your steps sound good. A third of the project size does not speak for those packages anyway :D

I don't have much time at the moment, so feel free to do it yourself

Fyi, I'm planning to have a look at API code generation in the next weeks and I started a native android app - mainly because I have no experience and wanted to try it out, not sure if this gets anywhere. If so I will share it :)

Issue #224, opened 2022-09-01
Move the most recent value in sats of an order calculation to the backend


KoalaSat:
**Is your feature request related to a problem? Please describe.**
As discussed in https://github.com/Reckless-Satoshi/robosats/pull/219 we want to move the most recent value in sats of an order calculation to the backend.

**Describe the solution you'd like**
I found these 2:

https://github.com/Reckless-Satoshi/robosats/blob/eff58dc91d93fe98a6738e0e3898ed2a476d2ebc/api/models.py#L382
https://github.com/Reckless-Satoshi/robosats/blob/eff58dc91d93fe98a6738e0e3898ed2a476d2ebc/api/models.py#L390

but  the only logic for `t0_satoshis` is this:

https://github.com/Reckless-Satoshi/robosats/blob/eff58dc91d93fe98a6738e0e3898ed2a476d2ebc/api/views.py#L146

Looks like there was an original intention to keep the sats on creation and the sats right now, that's cool, but at some point it got lost and `last_satoshis` was used as `t0_satoshis`.

I like the original idea, I had a look and there are not too many places where `last_satoshis` is being used

https://github.com/Reckless-Satoshi/robosats/search?q=last_satoshis

Is it a good idea to try a renaming? so we use `t0_satoshis` everywhere and start calculating `last_satoshis` on every request?

**Describe alternatives you've considered**
We might just create a new parameter and leave the other params as they are.

**Additional context**
I would create 2 separate PR. First one that fix the naming. And a second one that actually adds the new calculation and expose it on the endpoint.

Reckless-Satoshi:
Indeed, we keep track of t0_satoshis (simply noting them down). It might be truem that after that, t0_satoshis sees no use, altough, their intended use was to for validation that the order is not larger than the max order allowed (we might have used `last_satoshis` as it made no difference)

`last_satoshis` is only updated in the few requests that need them to be updated at the moment. That is: when the order is created and when the order is taken and the bond locked. It makes sense we might not want to update this value anymore after the order is taken. Therefore, for simplicity, I also think it might be best to create one more variable that we keep updated on every request, it can use the same function `Logics.satoshis_now()`.

KoalaSat:
I have been revisiting this proposal, and I'm staring to consider I was wrong about the need of having `satoshis_now`, as you asked, there is no need to have the amount of satoshis now, since it's not affected by the actual price but by the amount at the asking price. 

[...]

PR #226, opened 2022-09-06
 Move the most recent value in sats of an order calculation to the backend

KoalaSat:
Fixes: https://github.com/Reckless-Satoshi/robosats/issues/224

The value calculated by 

https://github.com/Reckless-Satoshi/robosats/blob/main/frontend/src/components/Charts/DepthChart/index.tsx#L84

Was already being calculated on the back-end, this PR exposes that value so it can be used for the depth chart generation.


api/views.py
```diff
@@ -268,6 +268,7 @@ def get(self, request, format=None):
         data["is_fiat_sent"] = order.is_fiat_sent
         data["is_disputed"] = order.is_disputed
         data["ur_nick"] = request.user.username
+        data["satoshis_now"] = Logics.satoshis_now(order)
```
Reckless-Satoshi, line 271: This will add the attribute `satoshis_now` to the responses of the endpoint `/api/order/`  but not to the responses of `/api/book/` . It should be added also somewhere around Line 854 for this attribute to be sent once the list of orders is requested in BookView( ) https://github.com/Reckless-Satoshi/robosats/blob/3eeff8c142d11e776b327fe0d9f2495d4acca86a/api/views.py#L854

KoalaSat: I see, makes sense, I think I went too fast instead of installing the full platform and test the back-end 😞

KoalaSat:
@Reckless-Satoshi Everything should be ready to be tested, I also included the fix for the max range when you use the price

frontend/src/components/BookPage.js
```diff
@@ -262,16 +262,21 @@ class BookPage extends Component {
...
-              </Tooltip>
-            )} 
+            renderCell: (params) => {
+              const currencyCode = this.getCurrencyCode(params.row.currency)
+              return (
+                <Tooltip placement="left" enterTouchDelay={0} title={pn(params.row.price) + " " + currencyCode + "/BTC" }>
+                <div style={{ cursor: "pointer" }}>{parseFloat(parseFloat(params.row.premium).toFixed(4))+"%" }</div>
+                </Tooltip>
+              )
+            }
```
KoalaSat, line 279: This fixes the currency code not being displayed on mobile

frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -73,19 +74,22 @@ const DepthChart: React.FC<DepthChartProps> = ({
...
-      setXRange(1500)
+
+      const medianValue = ~~median(prices)
+      const maxValue = prices.sort((a, b) => b - a).slice(0, 1)[0] || 1500
+      const maxRange = maxValue - medianValue
+      const rangeSteps = maxRange / 10
+
+      setCenter(medianValue)
+      setXRange(maxRange)
+      setRangeSteps(rangeSteps)
```
KoalaSat, line 85: This fixes the max range not being calculated for other currencies

Issue #195, opened 2022-07-30
Adapt every UI component to auto scale in size


Reckless-Satoshi:
**Is your feature request related to a problem? Please describe.**
Most MUI components we are using can be scale in size up or down simply by changing the value of `theme.typography.fontSize`. However, the frontend currently has many hardcoded `width` and `height` props that we have implemented. If these were also relative to `fontSize`, it would trivial to adjust the size of the app, highly increasing the usability on phones with small screens. This enhancement will also help us reuse a few of the components in RoboSats PRO frontend (it will have a default smaller fontSize in order to pack more data in the screen).

**Describe the solution you'd like**
The default `theme.typography.fontSize` is 14, so for example if we change it to 12 the app will be zoomed out to ~85%.
https://github.com/Reckless-Satoshi/robosats/blob/aad87e7d98be30bbbb814b91534e0709724c2bc4/frontend/src/components/App.js#L26
```
lightTheme = createMuiTheme({
  typography: {
    fontSize: 12,
  },
});
```
However, when changing the size of elements this way, many of the hard-coded icons/images/boxes will remain the same size (... bad practices in our initial implementation). The result is a clunky looking frontend. 

It would be great to make all of the hardcoded values automatically scale relative to `props.theme.typography.fontSize`. For that we will have to use `withTheme` in the class components and `useTheme` in the functional components. In addition, some of these values are directly hard-coded in `static/index.css` (for example the size of some Robot Avatars), ideally we should only rely on MUI theming.

This is a simple task, yet there might be dozens of hard-coded sizes all throughout the frontend. It will take reviewing and editing almost every file in `/frontend/src/components` then testing that the changes do actually lead to scalable components.

**Additional context**
This task is ⚡**rewarded with 300K Sats**⚡ . Drop a comment if you want to be assigned.

fmitjans:
I would like to work on this task

fmitjans:
Hi, I ran into a problem trying to use `withTheme` and `useTheme`.  Importing any of these functions from the `@mui/material/styles` module will result in the following error:

`MUI: withTheme is no longer exported from @mui/material/styles.`
`You have to import it from @mui/styles.`
`See https://mui.com/r/migration-v4/#mui-material-styles for more details.`

However, `@mui/styles`  is incompatible with React 18, as explained here: https://github.com/mui/material-ui/issues/33066#issuecomment-1149932811

[...]

PR #233, opened 2022-09-09
Current progress on issue #195 (auto-scale to fontsize)

fmitjans:
As discussed on the telegram group, this is my partial progress for issue #195.
I managed to merge and rebase my changes over the latest [main](https://github.com/Reckless-Satoshi/robosats/tree/main) commit, and everything seems to work fine when running the frontend.

One small issue: I noticed the Expiry Timers section on page /make > customize doesn't show the option to pick times*. 
The latest [main](https://github.com/Reckless-Satoshi/robosats/tree/main) commit also had this problem, and my branch showed it just fine (branching from commit 7fb2767), so this is probably caused by one of the latest commits. I thought it would still be better to rebase my changes to the latest commit instead of an old one.

*Specifically: this section is not showing:

https://github.com/Reckless-Satoshi/robosats/blob/645671265ab19d3fd127fa08f21dd76e9ebfa423/frontend/src/components/MakerPage.js#L601-L665

### DIFF
--- a/frontend/src/components/HomePage.js
+++ b/frontend/src/components/HomePage.js
@@ -40,9 +40,11 @@ export default class HomePage extends Component {
     }
 
     render() {
+        const fontSize = this.props.theme.typography.fontSize;
+        const fontSizeFactor = fontSize / 14; // default fontSize is 14
         return (
               <Router >
-                  <div className='appCenter'>
+                  <div className='appCenter' style={{height: window.innerHeight - 40*fontSizeFactor - 20, overflow:'auto'}}>
                     <Switch>
                         <Route exact path='/' render={(props) => <UserGenPage {...props} {...this.state} {...this.props} setAppState={this.setAppState}/>}/>
                         <Route path='/ref/:refCode' render={(props) => <UserGenPage {...props} {...this.state} {...this.props} setAppState={this.setAppState}/>}/>
@@ -51,8 +53,8 @@ export default class HomePage extends Component {
                         <Route path="/order/:orderId" render={(props) => <OrderPage {...props} {...this.state} {...this.props} setAppState={this.setAppState}/>}/>
                     </Switch>
                   </div>
-                  <div className='bottomBar'>
-                    <BottomBar redirectTo={this.redirectTo} {...this.state} setAppState={this.setAppState} />
+                  <div className='bottomBar' style={{height: `${40*fontSizeFactor}px`, width: window.innerWidth}}>
+                    <BottomBar redirectTo={this.redirectTo} {...this.state} {...this.props} setAppState={this.setAppState} />
                   </div>
               </Router>
           );
--- a/frontend/src/components/HomePage.js
+++ b/frontend/src/components/HomePage.js
@@ -44,11 +44,11 @@ export default class HomePage extends Component {
               <Router >
                   <div className='appCenter'>
                     <Switch>
-                        <Route exact path='/' render={(props) => <UserGenPage {...props} {...this.state} setAppState={this.setAppState}/>}/>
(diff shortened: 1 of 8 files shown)


frontend/src/components/HomePage.js
```diff
@@ -40,19 +40,21 @@ export default class HomePage extends Component {
     }
 
     render() {
+        const fontSize = this.props.theme.typography.fontSize;
+        const fontSizeFactor = fontSize / 14; // default fontSize is 14
         return (
               <Router >
-                  <div className='appCenter'>
+                  <div className='appCenter' style={{height: window.innerHeight - 40*fontSizeFactor - 20, overflow:'auto'}}>
```
Reckless-Satoshi, line 47: Breaks the UI as the app is not vertically centered and the height is too tight (shows annoying scroll bar on the middle right side)

Reckless-Satoshi:
Amazing! Thanks for this. I am temporarily hiding again the zoom-in and zoom-out buttons on the main branch. I also am restoring the original AutocompletePayments.js, as I do not understand well why it looks different.

Good catch for the Expiry Timers. I guess some dependency got lost (yet it does not spell any error... weird!). Fixed it before merge.

I noticed the maker page does not really scale in size. Do you know why? Also noticed many sizes in "em" units instead of "px" does using "em" solves the austcaling issue without need to use fontSize theme variable?

This work counts for ~half the reward on the Dev Rewards panel. Issue the invoice with the final PR :D

fmitjans:
> does using "em" solves the autoscaling issue without need to use fontSize theme variable?

Yes, I realized that the 'em' unit changes with the page font size (unlike the fixed 'px' unit), so I figured it would be a better solution than passing fontSize as a prop, which increases the code's complexity, specially with nested components. I checked with the element inspector and it has the same effect as the original solution

> I noticed the maker page does not really scale in size. Do you know why?

I'm not sure what you mean. It does scale in a weird way, because many elements in the maker page had fixed width and unspecified height, so I only changed their width to a relative one (using 'em'), as a result the page scales more to the sides than in height. If this is what you mean, I can specify every size so that the page scales as if zoomed, it shouldn't be hard.
The page should definitely scale somewhat, if it doesn't change at least a little there may be an error.

>  I also am restoring the original AutocompletePayments.js, as I do not understand well why it looks different.

The AutoCompletePayments component is part of the maker page. Before the changes it didn't change in sizes along with the rest of the page, so for example, if you zoomed in, you can see the list moving gradually to the left. In any case, this effect is small and it might be difficult to notice without the page inspector (if you inspect the div or the svgs in the payments autocomplete section, you can see the size stays fixed while the rest of the page changes, this doesn't happen when the commit is applied).
The page still works fine without these changes though, I'll leave the decision to you to keep them or not

Reckless-Satoshi:
> I checked with the element inspector and it has the same effect as the original solution
Awesome!

> I'm not sure what you mean. It does scale in a weird way, because many elements in the maker page had fixed width and unspecified height, so I only changed their width to a relative one (using 'em'), as a result the page scales more to the sides than in height. If this is what you mean, I can specify every size so that the page scales as if zoomed, it shouldn't be hard.

When I tested it, the makerpage was not scaling at all. But I discovered that it had to do with using the Dark theme. I fixed the conflict between light/dark theme and the fontSize in App.js in one of the latest commit.

> I felt like it would be better to change everything to relative sizes now rather than leave it hardcoded
The page still works fine without these changes though, I'll leave the decision to you to keep them or not

If using "em" instead of pixels is a general solution, let's just use em as unit and write all sizes in "em".

Since the makerpage did not zoom in or out for me, the only thing weird I noticed is the AutocompletePayments was less wide than usual.

PR #235, opened 2022-09-09
Move the most recent value in sats of an order calculation to the backend

KoalaSat:
Fixes: https://github.com/Reckless-Satoshi/robosats/issues/224

The value calculated by

https://github.com/Reckless-Satoshi/robosats/blob/main/frontend/src/components/Charts/DepthChart/index.tsx#L84

Was already being calculated on the back-end, this PR exposes that value so it can be used for the depth chart generation.

Looks like updating the fork removed the original PR https://github.com/Reckless-Satoshi/robosats/pull/226

### DIFF
--- a/frontend/src/components/BookPage.js
+++ b/frontend/src/components/BookPage.js
@@ -246,9 +246,10 @@ class BookPage extends Component {
               type: 'number',
               width: 140,
               renderCell: (params) => {
+                const currencyCode = this.getCurrencyCode(params.row.currency);
                 return (
                   <div style={{ cursor: 'pointer' }}>
-                    {pn(params.row.price) + ' ' + params.row.currency + '/BTC'}
+                    {pn(params.row.price) + ' ' + currencyCode + '/BTC'}
                   </div>
                 );
               },
--- a/frontend/src/components/Charts/DepthChart/index.tsx
+++ b/frontend/src/components/Charts/DepthChart/index.tsx
@@ -54,6 +54,7 @@ const DepthChart: React.FC<DepthChartProps> = ({
   const theme = useTheme();
   const [enrichedOrders, setEnrichedOrders] = useState<Order[]>([]);
   const [series, setSeries] = useState<Serie[]>([]);
+  const [rangeSteps, setRangeSteps] = useState<number>(8);
   const [xRange, setXRange] = useState<number>(8);
   const [xType, setXType] = useState<string>('premium');
   const [currencyCode, setCurrencyCode] = useState<number>(1);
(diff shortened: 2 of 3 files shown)


frontend/src/components/Charts/DepthChart/index.tsx
```diff
@@ -73,19 +74,22 @@ const DepthChart: React.FC<DepthChartProps> = ({
...
-      setXRange(1500)
+
+      const medianValue = ~~median(prices)
+      const maxValue = prices.sort((a, b) => b - a).slice(0, 1)[0] || 1500
+      const maxRange = maxValue - medianValue
+      const rangeSteps = maxRange / 10
+
+      setCenter(medianValue)
+      setXRange(maxRange)
+      setRangeSteps(rangeSteps)
```
KoalaSat, line 85: This fixes the max range not being calculated for other currencies

frontend/src/components/BookPage.js
```diff
@@ -262,16 +262,21 @@ class BookPage extends Component {
...
-              </Tooltip>
-            )} 
+            renderCell: (params) => {
+              const currencyCode = this.getCurrencyCode(params.row.currency)
+              return (
+                <Tooltip placement="left" enterTouchDelay={0} title={pn(params.row.price) + " " + currencyCode + "/BTC" }>
+                <div style={{ cursor: "pointer" }}>{parseFloat(parseFloat(params.row.premium).toFixed(4))+"%" }</div>
+                </Tooltip>
+              )
+            }
```
KoalaSat, line 279: This fixes the currency code not being displayed on mobile

Reckless-Satoshi:
Just tested it locally and created 4 orders. The API calls works, each book order has satoshis_now. However the DepthChart does not appear (the spinner stays there). I cannot say whether this is an issue with this PR or of having too few orders since it is the first time I test the DepthChart on a local environment. 

It might be best if you manage to run the stack locally as well, unfortunately this feature cannot be tested with the production backend. Another possible route is to split the PR in two. We first add satoshis_now to the backend, it will be safe to update the mainnet platform, then the frontend depthchart can be tested by itself.

KoalaSat:
> Just tested it locally and created 4 orders. The API calls works, each book order has satoshis_now. However the DepthChart does not appear (the spinner stays there). I cannot say whether this is an issue with this PR or of having too few orders since it is the first time I test the DepthChart on a local environment.
> 
> It might be best if you manage to run the stack locally as well, unfortunately this feature cannot be tested with the production backend. Another possible route is to split the PR in two. We first add satoshis_now to the backend, it will be safe to update the mainnet platform, then the frontend depthchart can be tested by itself.

Let's go for the 2nd option, I'll create the PR

KoalaSat:
https://github.com/Reckless-Satoshi/robosats/pull/236

KoalaSat:
Found the problem with tesnet, it was caused by a very specific bug caused when the `lastDayPremium` was zero. Fixed now.
image

PR #239, opened 2022-09-11
Fix error when query parameters not passed

redphix:
if no parameters are passed to `/api/book`, then just return orders of ALL currency and of ANY type.

Currently, if no query parameters are passed, then the API returns 500 status code.


Reckless-Satoshi:
Tested and works as intended. 

Indeed, it was annoying to have to send those two parameters to get all book orders. Now that endpoint is more intuitive.

Congrats on your first contribution to RoboSats.

PR #170, opened 2022-06-25
Mobile react-native Webview app

Reckless-Satoshi:
React Native App with simple react-native-webview and react-native-tor over typescript template. As described in https://github.com/Reckless-Satoshi/robosats/issues/42

- [x] Create setup guide
- [x] Initialize over typescript template
- [x] Webview app over clearnet tests
- [x] Load index/css/main.js from local (file://)
- [ ] Add react-native-tor (seems tough, not well maintained, might require rewritting all calls and passing them onMessage between react-native and webview JS app)
- [ ] Implement push notifications
- [x] Icons and styling

PR #242, opened 2022-09-13
Frontend API client

KoalaSat:
This PR is the first step for https://github.com/Reckless-Satoshi/robosats/pull/170 , it creates an abstraction layer for the API requests (for this generic case `ApiWebClient`), so it can be easily replaced in future steps by other methods to communicate with the back-end (ex: `ApiNativeClient`).

I prefer to leave web-sockets as a totally different task and first validate this approach.

This change is huge and requires deep testing.

### DIFF
--- a/frontend/src/components/UserGenPage.js
+++ b/frontend/src/components/UserGenPage.js
@@ -27,6 +27,7 @@ import { genKey } from '../utils/pgp';
 import { getCookie, writeCookie, deleteCookie } from '../utils/cookies';
 import { saveAsJson } from '../utils/saveFile';
 import { copyToClipboard } from '../utils/clipboard';
+import { apiClient } from '../services/api/index';
 
 class UserGenPage extends Component {
   constructor(props) {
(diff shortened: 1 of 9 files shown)


KoalaSat:
I tested all endpoints and complete an entire order, everything seems to work, but would be great if someone double checks

frontend/src/components/UserGenPage.js
```diff
@@ -63,27 +64,22 @@ class UserGenPage extends Component {
...
+        encrypted_private_key: key.encryptedPrivateKeyArmored,
+        unique_values: strength.uniqueValues,
+        counts: strength.counts,
+        length: token.length,
+        ref_code: refCode,
       };
     });
 
-    console.log(requestOptions);
+    console.log(requestBody);
```
Reckless-Satoshi, line 79: I think we can use this change to get rid of this legacy `console.log()`

frontend/src/services/api/ApiWebClient/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    };
+    return fetch(path, requestOptions).then((response) => response.json())
+  }
+
+  public delete: (path: string) => Promise<object> = (path) => {
+    const requestOptions = {
+      method: 'DELETE',
+      headers: this.getHeaders(),
+    };
+    return fetch(path).then((response) => response.json())
```
Reckless-Satoshi, line 32: `fetch(path)` is missing `requestOptions`, so this DELETE request behaves as GET. Robot deletion fails with 500 Error.

Reckless-Satoshi (commented):
Looks neat an clean. I tested slightly but still need to fully test an end-to-end trade still to make sure we do not merge a broken frontend.

frontend/src/services/api/ApiWebClient/index.ts
KoalaSat: changed, good catch!

Reckless-Satoshi:
Working as expected after a few full trades. Merging!

PR #244, opened 2022-09-14
Enable Telegram notifications for order takers

Reckless-Satoshi:
Feature request, allow order takers to enable Telegram Notifications as well. It is a bad privacy trade-off for the functionality it adds, but hey, it keeps getting requested.

In addition, we move the "Enable Telegram" button to the robot profile to improve UX.

- [x] Create EnableTelegramDialog functional component.
- [x] Add Telegram messages.py for takers.
- [x] Dual message (maker and taker) on every call.
- [x] Serve Telegram activation token with `/info/` if user is logged in.
- [x] Move Telegram Activation button from the published order stage on the `TradeBox` component to the `Profile` Dialog (watch out! A Dialog that opens a Dialog, things can get weird...)

### DIFF
--- a/api/logics.py
+++ b/api/logics.py
@@ -221,7 +221,6 @@ class Logics:
             order.expires_at = timezone.now() + timedelta(
                 seconds=order.t_to_expire(Order.Status.TAK))
             order.save()
-            # send_message.delay(order.id,'order_taken') # Too spammy
             return True, None
 
     def is_buyer(order, user):
@@ -315,7 +314,6 @@ class Logics:
         elif order.status == Order.Status.TAK:
             cls.cancel_bond(order.taker_bond)
             cls.kick_taker(order)
-            # send_message.delay(order.id,'taker_expired_b4bond') # Too spammy
             return True
 
         elif order.status == Order.Status.WF2:
@@ -882,7 +880,6 @@ class Logics:
             # adds a timeout penalty
             cls.cancel_bond(order.taker_bond)
             cls.kick_taker(order)
-            # send_message.delay(order.id,'taker_canceled_b4bond') # too spammy
             return True, None
 
             # 4) When taker or maker cancel after bond (before escrow)
--- a/api/messages.py
+++ b/api/messages.py
@@ -8,6 +8,7 @@ class Telegram():
     ''' Simple telegram messages by requesting to API'''
 
     session = get_session()
+    site = config('HOST_NAME')
 
     def get_context(user):
         """returns context needed to enable TG notifications"""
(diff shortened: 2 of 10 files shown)


api/messages.py
```diff
@@ -41,188 +42,131 @@ def send_message(self, user, text):
...
-        queryset = Order.objects.filter(maker=user)
-        order = queryset.last()
-
-        print(str(order.id))
         if lang == 'es':
-            text = f'Hola {user.username}, te enviaré un mensaje cuando tu orden con ID {str(order.id)} haya sido tomada.'
+            text = f'Hola {user.username}, te enviaré notificaciones sobre tus órdenes en RoboSats.'
         else:
-            text = f"Hey {user.username}, I will send you a message when someone takes your order with ID {str(order.id)}."
+            text = f"Hey {user.username}, I will send you notifications about your RoboSats orders."
```
KoalaSat, line 53: Maybe not for this PR but probably it's worth it to include also i18n on the backend https://github.com/danhper/python-i18n

frontend/src/components/Dialogs/EnableTelegram.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <Dialog
+      open={open}
+      onClose={onClose}
+      aria-labelledby='enable-telegram-dialog-title'
+      aria-describedby='enable-telegram-dialog-description'
+    >
+      <DialogTitle id='open-dispute-dialog-title'>{t('Enable TG Notifications')}</DialogTitle>
+      <DialogContent>
+        <div style={{ textAlign: 'center' }}>
+          <QRCode
```
KoalaSat, line 42: Is it not possible to also include on click a redirection to bot's web page same as below?

frontend/src/components/TradeBox.js
```diff
@@ -1409,7 +1335,7 @@ class TradeBox extends Component {
         enterTouchDelay={0}
         title={
           <Trans i18nKey='open_dispute'>
-            To open a dispute you need to wait{' '}
+            To open a dispute you need to wait
```
KoalaSat, line 1338: Not a translation?

api/logics.py
```diff
@@ -882,7 +880,6 @@ def cancel_order(cls, order, user, state=None):
             # adds a timeout penalty
             cls.cancel_bond(order.taker_bond)
             cls.kick_taker(order)
-            # send_message.delay(order.id,'taker_canceled_b4bond') # too spammy
```
KoalaSat, line 885: I'm not a fan of leaving commented code, do you want to just check how it goes?

api/messages.py
Reckless-Satoshi: Yes, agree. There is no internationalization implemented on the backend yet, but would be very useful for the Telegram messages.

frontend/src/components/TradeBox.js
Reckless-Satoshi: It is a translation, in this case we <Trans/> component instead of t(). We use the <Trans/> here because this translation has a children component <Countdown...>.

When running `npm run format` a few days ago the `{' '}` was introduced, I am not sure why. It made the <Countdown> component dissapear from the translations, as they were tagged with <1></1> (that used signal where to render the <Coundown/> in the translation) move and instead render the empty string ' '. 
https://github.com/Reckless-Satoshi/robosats/blob/cd8b607891a786c77986ffee45dbd75f51610a01/frontend/static/locales/en.json#L461

frontend/src/components/Dialogs/EnableTelegram.tsx
Reckless-Satoshi: Do you mean open the `t.me` website also on click on the <QRcode />? Sounds good!

api/logics.py
Reckless-Satoshi: Agree,

I had commented out this line long time ago. I was not sure the solution we found was good enough. But now that I revisited this file, I think it's clear the solution worked out and it is time to delete this commented code, forever (actually, this PR also deletes the function this line is calling to) :D

Reckless-Satoshi:
Thanks for your time reviewing @KoalaSat !

frontend/src/components/TradeBox.js
Reckless-Satoshi: (sorry for mixing this bug fix in this PR... totally unrelated, it was bugging me!)

PR #245, opened 2022-09-14
Fix logic of checking whether non-participants can view order details

redphix:
Not sure if this is a typo or am I missing something(?)


Reckless-Satoshi:
It's a typo/bug, nice find.
Added a PR with one more commit to your branch to polish this patch (any known workarounds so I can add commits directly to a PR?)

redphix:
I have enabled this feature, but not sure how it works though. My guess is that you get the permission to clone my repo and you get write access only for that one branch(?) not sure, as never used this.
image

PR #246, opened 2022-09-16
Update 02-robots.md

okada-blue:
Include additional, pertinent information for hopefully greater clarity. Let me know if there is a better way to go about these small edits.


Reckless-Satoshi:
Looks good!

PR #251, opened 2022-09-21
Refactor avatars

KoalaSat:
This PR is another step for https://github.com/Reckless-Satoshi/robosats/pull/170. I have refactored and extracted to a single component all Avatar references, so the image file can me easily managed on Android's Tor request. 

Additionally it brings a great normalization to Avatars.

No changes should apply, with the exception of the main Avatar's status badge on the bottom bar, where it was intentionally bigger than in other parts of the app, Let me know what you think about it.
### Before
image
### After
image


frontend/src/components/Robots/RobotAvatar/index.tsx
```diff
@@ -29,45 +40,53 @@ const RobotAvatar: React.FC<DepthChartProps> = ({ order }) => {
...
-      <Badge variant='dot' overlap='circular' badgeContent='' color={statusBadgeColor()}>
+    if (orderType !== undefined) {
+      component = (
         <Badge
           overlap='circular'
           anchorOrigin={{ horizontal: 'right', vertical: 'bottom' }}
           badgeContent={statusBadge}
         >
-          <div style={{ width: 45, height: 45 }}>
-            <SmoothImage
```
Reckless-Satoshi, line 56: I wonder whether we can keep using `SmoothImage`, at least for the Order Book avatars. It looks a bit ugly when so many images are loading slowly from top to bottom. With `SmoothImage` they fade-in after fully loaded.
bookloads

frontend/src/components/OrderPage.js
```diff
@@ -757,50 +724,13 @@ class OrderPage extends Component {
...
-                                    {this.state.type ? (
-                                      <SendReceiveIcon
-                                        sx={{ height: '18px', width: '18px' }}
-                                        color='secondary'
-                                      />
-                                    ) : (
-                                      <SendReceiveIcon
-                                        sx={{
-                                          transform: 'scaleX(-1)',
-                                          height: '18px',
```
Reckless-Satoshi, line 782: Just a small detail has changed here, not too important. The badge with the arrows that indicates whether the user is selling/buying (arrows go away from the avatar if he sells, or go towards the avatar if he buys) is located on the left size of the avatar when it is the taker avatar in the Order page. Visually it helps to understand where the Sats leave and where they go.
How it looks on this PR:
onPR
How it used to look:
before

frontend/src/components/EncryptedChat.js
```diff
@@ -270,23 +271,11 @@ class Chat extends Component {
...
-                src={
-                  window.location.origin +
-                  '/static/assets/avatars/' +
-                  props.message.userNick +
-                  '.png'
-                }
-              />
-            </Badge>
+            <RobotAvatar
+              avatarClass="phoneFlippedSmallAvatar"
```
Reckless-Satoshi, line 275: Not sure if intentional but `className="phoneFlippedSmallAvatar"` is the button Profile button for mobile. It's an avatar modificed to be more of a "button". You have used it here instead of `className='flippedSmallAvatar'`. Now chat avatars look like this in both desktop and mobile:
chat
They used to look like this:
chat2

frontend/src/components/BottomBar.js
```diff
@@ -512,29 +496,13 @@ class BottomBar extends Component {
...
-                          ? window.location.origin +
-                            '/static/assets/avatars/' +
-                            this.props.nickname +
-                            '.png'
-                          : null
-                      }
-                    />
-                  </Badge>
+                  <RobotAvatar
+                    style={{ width: 55, height: 55 }}
```
Reckless-Satoshi, line 500: Once again here the image size seems to not be taken. The profile button size on mobile is now the size of all other avatars:
bottomavatar2

Used to look like this (55 x 55):
bottomavatar1

Reckless-Satoshi (commented):
Hey @KoalaSat 

This PR looks super. Simplifies and unifies the use of RobotAvatars all over the UI. The RobotAvatar component will, in addition, be very handy once we start working on the RoboSats PRO frontend.

There are a few minor aesthetic aspects that have changed with this PR. Would be great to fix the avatar sizes so they stick to the size given with the `style` prop, and have `SmoothImage` (or a similar behaviour) in the book.

Thanks a lot, awesome work!

frontend/src/components/BottomBar.js
KoalaSat: That's weird for me it doesn't look like that, what's your browser??

frontend/src/components/OrderPage.js
KoalaSat: But on the Book the arrow is on the right, right? I have on mind to try to normalize everything so all avatars looks identical so shape and icons becomes familiar for the user. What do you think?

frontend/src/components/Robots/RobotAvatar/index.tsx
KoalaSat: Let me try to find out if we can do something with the avatar component

frontend/src/components/EncryptedChat.js
KoalaSat: Yeah that was a typo

frontend/src/components/BottomBar.js
KoalaSat: Anyways I think it's about the same topic, my intention is to normalize everything not only form the tech side but also UX

frontend/src/components/Dialogs/Profile.tsx
```diff
@@ -167,13 +168,10 @@ const ProfileDialog = ({
...
-              <Avatar
-                className='profileAvatar'
-                sx={{ width: 65, height: 65 }}
-                alt={nickname}
-                src={
-                  nickname ? `${window.location.origin}/static/assets/avatars/${nickname}.png` : ''
-                }
+              <RobotAvatar
+                avatarClass="profileAvatar"
+                style={{ width: 65, height: 65 }}
```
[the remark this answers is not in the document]
KoalaSat: Looks like on your browser is happening differently, can you send a screenshot and share what's your browser?

KoalaSat:
Thanks for the comments @Reckless-Satoshi , all changes applied now, expect for the circle on the bottom bar and the arrow on the right. I believe it's less confusing like that but please let me know what you think and I'll apply the modifications.

Reckless-Satoshi:
Merging as is. We'll have ample time in the future to figure out in what corner the arrows should be :rofl: 

Looking forward to see all your work come together for the Android app :smile:

Reckless-Satoshi:
@KoalaSat It is already merged, however I realized on mobile view of the Order Book the avatars do not render. Won't be able to move it into the mainnet production env until we solve this one (ideally the fade in of avatars too).

KoalaSat:
> @KoalaSat It is already merged, however I realized on mobile view of the Order Book the avatars do not render. Won't be able to move it into the mainnet production env until we solve this one (ideally the fade in of avatars too).

I'm so sorry 😞 I always forgot the mobile view is different code, here is the fix https://github.com/Reckless-Satoshi/robosats/pull/254/files

PR #254, opened 2022-09-23
Mobile Avatar Book Smooth Load

KoalaSat:
Enabling smooth avatar load on mobile.


Reckless-Satoshi:
Tested, works exactly as intended! :rocket:

PR #256, opened 2022-09-26
Book functional component

Reckless-Satoshi:
## What does this PR do?
Turns the book table into a standalone component that is fully responsive.

Size of the book table will adapt to screen. The default number of rows will adapt to fit. The columns shown are ordered by importance and will be displayed according to the size of the display.

This feature is key if we want to reuse the book component in the new landing page and as a viewport in RoboSats Pro. It will also allow to fit the information better into any Android device.

### New columns behavior
The premium column is now colored following some basic logic. Outstanding premiums for buy offers are colored purple and bold, outstanding premiums for sell offers are colored blue and bold.

Expiry column: shows time left until expiration as a public order. It displays a circular determinate progress bar that becomes orange and red as time gets close to expiration. It displays the time left in the center in hours.

Timer column: displays `escrow_duration` time in hours and minutes.

Satoshis now: on large displays, the current size of the order in Satoshis is show. The value display is "thousands" of satioshis `${satoshis_now/1000}K`

Order id: on very large displays the Order ID is also shown.

### To test
Depth chart is also responsive to window resizes. Let's test by resizing the window and checking the two views ( depth and list) render correctly on every display. Special attention to be put into the columns the List view is displaying, some `width` resolution might display a combination of columns that is not as informative as it could be (e.g. on small windows, sometimes the Payment methods dissapear, yet these are very important).

## Checklist before merging
- [x] If its a frontend feature, I have ran prettier (`cd frontend; npm run format`).

### DIFF
--- a/frontend/src/components/BookTable.tsx
+++ b/frontend/src/components/BookTable.tsx
@@ -20,49 +20,64 @@ import FlagWithProps from './FlagWithProps';
 import { pn, amountToString } from '../utils/prettyNumbers';
 import PaymentText from './PaymentText';
 import RobotAvatar from './Robots/RobotAvatar';
+import hexToRgb from '../utils/hexToRgb';
+import statusBadgeColor from '../utils/statusBadgeColor';
 
-function statusBadgeColor(status) {
-  if (status === 'Active') {
-    return 'success';
-  }
-  if (status === 'Seen recently') {
-    return 'warning';
-  }
-  if (status === 'Inactive') {
-    return 'error';
-  }
+interface Props {
+  loading: boolean;
+  orders: Order[];
+  type: number;
+  currency: number;
+  maxWidth: number;
+  maxHeight: number;
 }
 
-function hexToRgb(c: string) {
-  if (c.includes('rgb')) {
-    const vals = c.split('(')[1].split(')')[0];
-    return vals.split(',');
-  }
-  if (/^#([a-f0-9]{3}){1,2}$/.test(c)) {
-    if (c.length == 4) {
-      c = '#' + [c[1], c[1], c[2], c[2], c[3], c[3]].join('');
-    }
-    c = '0x' + c.substring(1);
-    return [(c >> 16) & 255, (c >> 8) & 255, c & 255];
-  }
-  return '';
-}
-
-function mixColors(baseColor: string, accentColor: string, point: number) {
-  const baseRGB = hexToRgb(baseColor);
-  const accentRGB = hexToRgb(accentColor);
-  const redDiff = accentRGB[0] - baseRGB[0];
-  const red = baseRGB[0] + redDiff * point;
-  const greenDiff = accentRGB[1] - baseRGB[1];
-  const green = baseRGB[1] + greenDiff * point;
-  const blueDiff = accentRGB[2] - baseRGB[2];
-  const blue = baseRGB[2] + blueDiff * point;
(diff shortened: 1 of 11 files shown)


frontend/src/components/Robots/RobotAvatar/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Avatar, Badge, Tooltip } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { SendReceiveIcon } from '../../Icons';
 
-interface DepthChartProps {
+interface Props {
```
KoalaSat, line 7: It's good to have always identified the props per component, but I don't have a strong opinion neither 👍

frontend/src/components/BookTable.tsx
```diff
@@ -0,0 +1,624 @@
...
+  const baseRGB = hexToRgb(baseColor);
+  const accentRGB = hexToRgb(accentColor);
+  const redDiff = accentRGB[0] - baseRGB[0];
+  const red = baseRGB[0] + redDiff * point;
+  const greenDiff = accentRGB[1] - baseRGB[1];
+  const green = baseRGB[1] + greenDiff * point;
+  const blueDiff = accentRGB[2] - baseRGB[2];
+  const blue = baseRGB[2] + blueDiff * point;
+  return `rgb(${Math.round(red)}, ${Math.round(green)}, ${Math.round(blue)}, ${0.7 + point * 0.3})`;
+}
```
KoalaSat, line 61: Maybe we can move these functions to utils?

frontend/src/components/BookTable.tsx
```diff
@@ -0,0 +1,624 @@
...
+    columnMenuFilter: t('Filter'),
+    columnMenuHideColumn: t('Hide'),
+    columnMenuUnsort: t('Unsort'),
+    columnMenuSortAsc: t('Sort by ASC'),
+    columnMenuSortDesc: t('Sort by DESC'),
+    columnHeaderFiltersLabel: t('Show filters'),
+    columnHeaderSortIconLabel: t('Sort'),
+    booleanCellTrueLabel: t('yes'),
+    booleanCellFalseLabel: t('no'),
+  };
```
KoalaSat, line 113: this constant might be just declared on the component, it's not necessary to have a function

frontend/src/components/BookTable.tsx
```diff
@@ -0,0 +1,624 @@
...
+  );
+  const height = defaultPageSize * verticalHeightRow + verticalHeightFrame;
+
+  const [pageSize, setPageSize] = useState(0);
+  const [useDefaultPageSize, setUseDefaultPageSize] = useState(true);
+  useEffect(() => {
+    if (useDefaultPageSize) {
+      setPageSize(defaultPageSize);
+    }
+  });
```
KoalaSat, line 154: }, []);

frontend/src/components/Robots/RobotAvatar/index.tsx
Reckless-Satoshi: I see, sounds good to me. I realized we are not being consistent across components. It was easier to change this one rather than all the other ones :smile:

frontend/src/components/BookTable.tsx
Reckless-Satoshi: For some reason this change makes the `pageSize` stay as 0.

frontend/src/components/BookTable.tsx
Reckless-Satoshi: Done!

PR #265, opened 2022-10-03
Add MakerForm functional component

Reckless-Satoshi:
## What does this PR do?
This PR turns the maker page into responsive functional component. The new maker form could be used in any location, including a Dialog over the OrderBook and in RoboSats Pro. In the future UX rework, the maker page will altogether disapear, so only the MakerForm as a Dialog will remain. We take the chance to un-mess one of the messiest pieces of code.

- [x] Turn `MakerPage` into a typed functional component.
- [x] Improve `AutoCompletePayments`
- [x] Merge `basic` and `customize` tabs into a single Grid.
- [x] Create `Maker` Dialog on BookPage
- [x] Use the new `Maker.Model` data for advance filtering of the `BookTable`
- [ ] ~~Turn `isExplicit` from "Satoshis" to "ExactPrice" (the user will specify the exact exchange rate, not the amount of Sats)~~
- [x] Delete `BondlessTakers` option
- [x] Add matchmaker
- [x] Add new filter order util
- [x] Refactor into submodules (`components/BookPage`, `components/MakerPage`, ...)

## Checklist before merging
- [x] If its a frontend feature, I have ran prettier (`cd frontend; npm run format`).


frontend/src/utils/filterOrders.ts
```diff
@@ -0,0 +1,77 @@
...
+        order.min_amount > filter.minAmount * (1 - filter.threshold)
+      );
+    } else {
+      return (
+        order.amount < filter.maxAmount * (1 + filter.threshold) &&
+        order.amount > filter.minAmount * (1 - filter.threshold)
+      );
+    }
+  }
+};
```
KoalaSat, line 60: I think that the final logic should be the same in all scenarios right? the only change is the source, not sure if I did well but maybe something like this might help for future maintenance

````
const filterByAmount = function (order: Order, filter: AmountFilter) {
    const filterMaxAmount = filter.amount != '' ? filter.amount : filter.maxAmount
    const filterMinAmount = filter.amount != '' ? filter.amount : filter.minAmount
    const orderMinAmount = order.amount === '' || order.amount === null ? order.min_amount : order.amount
    const orderMaxAmount = order.amount === '' || order.amount === null ? order.max_amount : order.amount

    return (
        orderMaxAmount < filterMaxAmount * (1 + filter.threshold) &&
        orderMinAmount > filterMinAmount * (1 - filter.threshold)
    );
};
````

KoalaSat (approved):

frontend/src/utils/filterOrders.ts
Reckless-Satoshi: This is certainly much better. My thinking was not clear and the code reflected it :sweat_smile: . Thanks a lot!

Reckless-Satoshi:
Merging and moving it to testnet as a release candidate for v0.2.2-alpha

PR #275, opened 2022-10-10
Add currencies ETB, GEL, and UGX which Yadio now supports

sgornick:
For testing, I did nothing more than a sanity check on the following to ensure the rates were current:

https://api.yadio.io/json/ETB
https://api.yadio.io/json/GEL
https://api.yadio.io/json/UGX


Reckless-Satoshi:
Tested and works as expected. 

Thanks a lot for adding these currencies! :rocket:

PR #286, opened 2022-10-15
Android style fixes

Reckless-Satoshi:
## What does this PR do?
Fix small Android theme issues.

Deletes the "Loading RoboSats" overlay as it is redundant with the Loading CSS animation while `main.js` loads.

Detect whether the device has dark mode on and sets the background color of `SafeAreaView` accordingly. This is done in order to avoid a white screen flash of ~0.5 secs while opening the app on a dark mode device (untested).

## Checklist before merging

### DIFF
--- a/mobile/App.tsx
+++ b/mobile/App.tsx
@@ -150,7 +150,6 @@ const App = () => {
         allowsBackForwardNavigationGestures={false}
         mediaPlaybackRequiresUserAction={false}
         allowsLinkPreview={false}
-        renderLoading={() => <Text>Loading RoboSats</Text>}
         onError={(syntheticEvent) => <Text>{syntheticEvent.type}</Text>}
       />
     </SafeAreaView>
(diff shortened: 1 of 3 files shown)


Reckless-Satoshi:
@KoalaSat I am hoping that by using `Appearance` we can set the background color of the `SafeAreaView` to avoid the white flash when starting up the app. However I have not been able to test myself whether the fix works as intended: the device I connect for debugging is stuck on "light" mode. And after building a release APK for the other device I use, it does not want to install it :sweat_smile:

Reckless-Satoshi:
`lint:fix` made a small change in `services/tor` . Reinstalling dependencies from `package.json` installed a few newer versions.

KoalaSat:
> @KoalaSat I am hoping that by using `Appearance` we can set the background color of the `SafeAreaView` to avoid the white flash when starting up the app. However I have not been able to test myself whether the fix works as intended: the device I connect for debugging is stuck on "light" mode. And after building a release APK for the other device I use, it does not want to install it 😅

Looks good!

KoalaSat (approved):

PR #291, opened 2022-10-20
Fixes to python auto-formatting

Reckless-Satoshi:
## What does this PR do?
Small fixes and an attempt to get understanding into the flake/black linting action.

## Checklist before merging
- [ ] If it's a frontend feature, I have ran prettier `cd frontend; npm run format`. If it's a mobile app feature I ran `cd mobile; npm run format`.

### DIFF
--- a/.github/workflows/py-linter.yml
+++ b/.github/workflows/py-linter.yml
@@ -1,14 +1,19 @@
-name: Python Lint
+name: Python Backend Lint
 
 on:
   push:
     branches:
       - main
-  pull_request:
+    paths:
+      - '**.py'
+  pull_request_target:
     branches:
       - main
     paths:
       - '**.py'
+
+permissions:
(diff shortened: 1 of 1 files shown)


Reckless-Satoshi:
@redphix 
What do you think about the behavior of these checks? I think these are useful and not too intrusive.

redphix:
> I added only 1 change that does require of a Black fix. It seems however Black did not pick it.

Strange. It should annotate the problems with a diff that would fix the problem. It also should have just added a commit fixing the problems, since it's your own branch, and you have permissions to push. It did when I tested in my own fork. 

> On the other hand, Flake8 left 606 annotations :smile: 

Sounds off. I had close to 1.1k lint errors by flake8

redphix:
Also why is flake8 not respecting .flake8 config file :thinking:

PR #293, opened 2022-10-20
Annoy black

redphix:
## What does this PR do?
annoy black

## Checklist before merging
- [ ] If it's a frontend feature, I have ran prettier `cd frontend; npm run format`. If it's a mobile app feature I ran `cd mobile; npm run format`.

### DIFF
--- a/api/logics.py
+++ b/api/logics.py
@@ -1,17 +1,16 @@
+import ast
+import math
 from datetime import timedelta
-from django.utils import timezone
-from api.lightning.node import LNNode
-from django.db.models import Q, Sum
-
-from api.models import OnchainPayment, Order, LNPayment, MarketTick, User, Currency
-from api.tasks import send_message
-from decouple import config
-from api.utils import validate_onchain_address
 
 import gnupg
+from decouple import config
+from django.db.models import Q, Sum
+from django.utils import timezone
 
-import math
-import ast
+from api.lightning.node import LNNode
+from api.models import Currency, LNPayment, MarketTick, OnchainPayment, Order, User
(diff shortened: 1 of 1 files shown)


redphix:
Okay nice. The flake8 lints work with annotations.

redphix:
@Reckless-Satoshi 
But the black auto-fix should only work when you either
1. Open a PR yourself
2. push to main

Reckless-Satoshi:
> looks fine for now. What do you think?

I think it's a way better set up than what we had before! :smile: Thank you very much for getting this done :rocket:

redphix:
Final test I guess -- merging this should run the auto-fix and commit it to main branch.

Reckless-Satoshi:
> Final test I guess -- merging this should run the auto-fix and commit it to main branch.

Merged. Auto-fix did not add a new commit, possibly because it also needs `on.pull_request` event as trigger. In any case, do we want auto fix commits? It is probably enough with PR checks. Also a bit scary the bot committing changes to main :D

redphix:
@Reckless-Satoshi Yes, I was also having second thoughts about it :D. A bot commit changes to main is definitely an uncomfortable feeling. I guess we don't autofix then -- commit annotations are more that enough. The auto-fixing part can be documented/enforced using pre-commit.

redphix:
@Reckless-Satoshi Oh the reason It didn't work is because of this - https://github.com/redphix/robosats/blob/1dea4e73b6410ca741c85a4b6c4b8ab2e2649bcb/.github/workflows/py-linter.yml#L36

PR #295, opened 2022-10-20
Conform to pep8 rules

redphix:
## What does this PR do?
Conforms to pep8 standards for all py files

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+max-line-length = 88
+extend-ignore = E501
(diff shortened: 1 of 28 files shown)


redphix:
ah nice I missed a few and the linter was able to catch it! Useful already!

redphix:
There are still 5 errors, I want you to check them and see that fixing them wont cause some side effect. I am not comfortable changing these ones, as I don't understand them fully -- there could be a reason why you might import something but not really use it in there.

Reckless-Satoshi:
> I will build a coordinator image based on this PR and give it a test run to confirm everything still working.

@redphix 
Nothing in the trade pipeline, celery tasks, or other backend processes seems broken. Still many things I can't test quickly. But it looks okay to me to be merged. Let's also fix those 5 errors left, then we can run further tests on the Testnet platform in case something escaped.

Issue #177, opened 2022-06-27
RoboSats PRO. Let's draft the dream app of large liquidity providers and advanced RoboSats users.


Reckless-Satoshi:
Advanced users and large liquidity providers need better tools: a panel/workspace where they track many robots and orders, functions to batch load/save order configurations, multiple robots at once, further order customization, live order premium editing, etc. All these features add further complexity to the current frontend and will be very confusing for new users, yet they are critical for RoboSats to grow into a healthy and liquid market. So... Let's create a frontend specifically for advance users!

**Let's fork RoboSats frontend into two: _RoboSats Basic_ and _RoboSats PRO_**

### _RoboSats basic_
_RoboSats basic_ (or simply "RoboSats") will stay as is currenctly, located in `/frontend/src/` . Some of the advanced features included recently in it can be deleted (for example, the _Customize_ tab on the `OrderPage.js`). RoboSats basic must remain clean and easy to use for new users. RoboSats basic will also be eventually available as a Webview app for Android (and iOS) currently WIP https://github.com/Reckless-Satoshi/robosats/pull/170

### _RoboSats PRO_
_RoboSats PRO_ will be a new frontend, designed for Desktop only. _RoboSats PRO_ is a re-factoring of the current app with an added extra level of client-side abstraction: **_a workspace_**. A workspace consists of many robot identities with their many orders. A workspace should be easy to export and import from the browser app into a file in the local disk. The objective is simple: a user should be able to track the current state of all of his robots and orders with a quick glance (no need to click/scroll.) and interact with the prompt of any of them.

**This issue is intended to discuss users expectations, define app requirements and create a rough draft of how RoboSats PRO must look.** 

### This feature _strongly_ eligible for [developer compensation (pilot program)](https://learn.robosats.com/contribute/code/#developer-compensation-pilot-program)

Reckless-Satoshi:
I will roughly describe what I have in mind: "a table, with a toolbar on top", not too different of how _Google Sheets_ looks.

On each row of the table there will be a Robot identity with an active (or past) order that is being tracked. Columns on this table can be something like ["robot_nickname","robot_token","has_order","order_id","currency","amount","amount SATS","premium","exchange_rate","order_state","current_contract_box_prompt", "etc"] (justfor illustrative purpose, I did not think deeply how this table should look!!). 

[...]

PR #296, opened 2022-10-20
Refactor and Init RoboSats PRO

Reckless-Satoshi:
## What does this PR do?
This PR has a large refactor of `/frontend/src/`

The objective is to draw a clear path for the future development of RoboSats Basic and the new RoboSats Pro (https://github.com/Reckless-Satoshi/robosats/issues/177)

- It strips down every component that strictly belong to Robosats Basic in the new `/src/basic` . Basic is organized it in "Pages" (e.g. `MakerPage`) and a `NavBar`. 
- It adds a new `/src/pro` app that is a "fork" of basic. RoboSats PRO is intended to be a [react-grid-layout](https://github.com/react-grid-layout/react-grid-layout) canvas with `Widgets` (wrappers of the components we have built and a few new ones)
- All components that are to be used in both Basic and Pro now live in `/src/components` . 

Both apps share the rest, including: same `package.json` and `package-lock.json` , same `index.js` and same `App.tsx` (Same, `i18n` loader, same `UnsafeAlert`, same themes, etc...). However the `Main.tsx` is replaced by WebPack building two different bundles (basic uses `/src/basic/Main.tsx` pro uses `/src/basic/Main.tsx`). These two results minimized: basic.js (so far known as main.js) and  `pro.js` .

A single coordinator will serve both `basic.js` and `main.js`. As of now the routes remain identical, except that `/pro` e.g. `robosats.com/pro` will make you land in a "index.html" that loads the `pro.js` bundle.

After a lot of thought on whether `pro` should be it's own standalone CreatReactApp I believe this approach is superior. It will lead to the lowest development and maintenance overhead by focusing the development purely in the Widgets and makign Basic and Pro simple wrappers. It leads to lower CI/CD footprint as well as easier to understand for users and coordinators. There is a few cons, but if this become important, we can always eject RoboSats PRO into a stand alone at a later stage.

## Checklist before merging


Reckless-Satoshi:
Merged via CLI. https://github.com/Reckless-Satoshi/robosats/commit/1dea4e73b6410ca741c85a4b6c4b8ab2e2649bcb

PR #299, opened 2022-10-21
Add react layout grid playground

Reckless-Satoshi:
## What does this PR do?
This is a playground with react-layout-grid . It adds some test Widgets (e.g. MakerForm) and gives a structure to a simple RoboSats PRO client.

## Checklist before merging


Reckless-Satoshi:
Merging and deploying so it is easier for Frontend developers to use the `/pro` route.

Renaming the `basic.js` bundle back to `main.js` to avoid compatibility issues with previous selfhosted node clients.

Issue #175, opened 2022-06-27
Add REST API endpoint for the trade chat (Turtle mode chat)


Reckless-Satoshi:
**Is your feature request related to a problem? Please describe.**
The Tor network is under attack for the last 15 days. When the network becomes very slow, the chat WebSockets is unstable and hardly hardly usable. However, real time chatting works great, is a good UX and must be kept. As an alternative to TOR and its recent slow downs, an I2P mirror of RoboSats is on the making (https://github.com/Reckless-Satoshi/robosats/commit/131fa0fb99df868bb6ae8d1f7f69ed0930566430 example I2P mirror for the dev site), yet, the websocket chat is also somewhat broken when using the I2P mirror (users can receive messages but not send: the cause of this issue is known).

**Describe the solution you'd like**
It would be great if users can send messages on a POST request and fetch new messages with a GET request. We can call this "_Turtle mode chat_". It should be possible to enable it on the EncryptedChat.js component and be fully toggleable (on/off) seamlessly with the WebSockets implementation. It could poll the server every 30s for new messages.

This feature also requires some re-writting of the current _chat/consumers.py_ . Since a message sent via REST has to be detected and sent via Websocket to the counterpart.

**Describe alternatives you've considered**
There isn't really any other alternative when it comes to increase the reliability of the chat communication.

**Additional context**
A REST API endpoint to chat with your counterpart will also make it very easy to implement third party trading apps that use the RoboSats API (no need to implement a WebSockets pipeline only for the chat).

The frontend could detect whether "i2p" is part of the domain. If it is, _Turtle mode chat_ can be enabled by default.

Reckless-Satoshi:
Since the I2P site websocket has been fixed and works flawlessly this task priority is now very low.

PR #311, opened 2022-11-01
Fix navbar refactor bugs on native app

KoalaSat:
## What does this PR do?
Fixes #[175](https://github.com/Reckless-Satoshi/robosats/issues/175)

## Checklist before merging


frontend/package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "frontend",
-  "version": "0.2.2",
+  "version": "0.2.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "frontend",
-      "version": "0.2.2",
+      "version": "0.2.3",
```
KoalaSat, line 9: I ran `npm install` and got this change 🤷‍♂️

frontend/src/components/TradeBox/EncryptedChat/MessageCard/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+const MessageCard: React.FC<Props> = ({ message, isTaker, userConnected }) => {
+  const [showPGP, setShowPGP] = useState<boolean>();
+  const { t } = useTranslation();
+  const theme = useTheme();
+
+  const takerCardColor = theme.palette.mode === 'light' ? '#d1e6fa' : '#082745';
+  const makerCardColor = theme.palette.mode === 'light' ? '#f2d5f6' : '#380d3f';
+  const cardColor = isTaker ? takerCardColor : makerCardColor;
```
KoalaSat, line 27: now messages backgrounds will match user's role

frontend/src/components/TradeBox/EncryptedChat/index.tsx
```diff
@@ -1,522 +1,56 @@
...
-  const [peerPubKey, setPeerPubKey] = useState<string>();
-  const [token] = useState<string>(systemClient.getCookie('robot_token') || '');
+const EncryptedChat: React.FC<Props> = ({
+  turtleMode,
+  orderId,
+  takerNick,
+  userNick,
+  chatOffset,
+}: Props): JSX.Element => {
   const [messages, setMessages] = useState<EncryptedChatMessage[]>([]);
```
KoalaSat, line 35: By keeping messages here, they we'll be shared between both components and switching  won't affect to the UX, if not the chant went totally blank because the component were loading entirely

frontend/src/components/TradeBox/EncryptedChat/index.tsx
```diff
@@ -1,522 +1,56 @@
...
+  turtleMode: boolean;
   orderId: number;
+  takerNick: string;
+  makerNick: string;
   userNick: string;
+  chatOffset: number;
 }
-
-interface EncryptedChatMessage {
+export interface EncryptedChatMessage {
```
KoalaSat, line 13: @Reckless-Satoshi I see on your PR you have other models, maybe we need to sync on this

Reckless-Satoshi: Will take care of it when that PR needs to be merged. This one will go first.

chat/views.py
```diff
@@ -79,7 +81,7 @@ def get(self, request, format=None):
             }
             messages.append(data)
 
-        response = {"peer_connected": peer_connected, "messages": messages}
+        response = {"peer_connected": peer_connected, "messages": messages, "peer_pubkey": peer_public_key}
```
Reckless-Satoshi, line 84: perfect!

frontend/src/components/TradeBox/EncryptedChat/EncryptedTrutleChat/index.tsx
```diff
@@ -0,0 +1,303 @@
+import React, { useEffect, useState } from 'react';
```
Reckless-Satoshi, line 1: Small typo on component folder name `Trutle`

frontend/src/components/TradeBox/index.js
```diff
@@ -72,6 +75,7 @@ class TradeBox extends Component {
       badInvoice: false,
       badAddress: false,
       badStatement: false,
+      turtle_mode: false,
```
Reckless-Satoshi, line 78: Out of curiosity, why not use camel case `turtleMode` here too? No fix needed as we are rewriting this component in parallel.

frontend/src/components/TradeBox/index.js
```diff
@@ -1438,6 +1442,29 @@ class TradeBox extends Component {
             </b>{' '}
             {' ' + this.stepXofY()}
           </Typography>
+          <Grid item>
```
Reckless-Satoshi, line 1445: If placed as the first item of `<Grid container ...>` and with `style={{position:'fixed', right:'1em'}}` the Switch would probably render on the top right corner of the TradeBox. No need to fix as this work would have to be re-done next week on the new TradeBox anyway.

frontend/static/locales/es.json
```diff
@@ -345,6 +345,7 @@
   "Messages": "Mensajes",
   "Verified signature by {{nickname}}": "Firma de {{nickname}} verificada",
   "Cannot verify signature of {{nickname}}": "No se pudo verificar la firma de {{nickname}}",
+  "Activate turtle mode (Use it when the connection is slow)": "Activar modo tortuga (Úsalo cuando tu conexión es lenta)",
```
Reckless-Satoshi, line 348: Perfect!

frontend/src/components/TradeBox/EncryptedChat/EncryptedTrutleChat/index.tsx
```diff
@@ -0,0 +1,303 @@
...
+  }, [serverMessages, peerPubKey]);
+
+  useEffect(() => {
+    if (chatOffset > lastIndex) {
+      loadMessages();
+    }
+  }, [chatOffset]);
+
+  const loadMessages: () => void = () => {
+    apiClient.get(`/api/chat?order_id=${orderId}&offset=${lastIndex}`).then((results: any) => {
```
Reckless-Satoshi, line 79: This call is not needed. The `chat` object already is in `this.props.data.chat` (it is returned with the `/api/order` call when the offset is supplied).

frontend/src/basic/OrderPage/index.js
```diff
@@ -109,6 +110,7 @@ class OrderPage extends Component {
       currencyCode: this.getCurrencyCode(newStateVars.currency),
       penalty: newStateVars.penalty, // in case penalty time has finished, it goes back to null
       invoice_expired: newStateVars.invoice_expired, // in case invoice had expired, it goes back to null when it is valid again
+      chat_offset: this.state.chat_offset + newStateVars?.chat?.messages.length,
```
Reckless-Satoshi, line 113: This grows the offset exponentially?

Chat offset must be the index of the last message received. 

It is 0 when the component is mount, then it updates to the index of the newest message in the state. After making the call `/api/order ... offset=...` the return contains all new messages. Those have to be merged with the existing messages. That's it ! :D 

We do not need extra API call for turlte mode. All we need for turtle mode is supply an extra param to `/api/order` (the offset).

Reckless-Satoshi: The component is not working. It seems the `offset` param is given to `/api/order` with a NAN value so the backend erros out with 500. Ideally we do not want to supply the `offset` param if we have not activated the turtle mode.

frontend/src/components/TradeBox/EncryptedChat/EncryptedTrutleChat/index.tsx
```diff
@@ -0,0 +1,303 @@
...
+interface Props {
+  orderId: number;
+  userNick: string;
+  takerNick: string;
+  chatOffset: number;
+  messages: EncryptedChatMessage[];
+  setMessages: (messages: EncryptedChatMessage[]) => void;
+}
+
+const EncryptedTurtleChat: React.FC<Props> = ({
```
Reckless-Satoshi, line 30: Very clean solution for the 2 modes and the re-utilization of header and bottom buttons!

frontend/src/components/TradeBox/index.js
KoalaSat: My brain was out I guess, I can change it anyways to avoid confusion

frontend/src/components/TradeBox/EncryptedChat/EncryptedTrutleChat/index.tsx
KoalaSat: It's hard to maintain the data coming from the parent and a good alignment between both turtle and websocket mode components, I did at first but didn't like at all. What if we keep this and just return a number instead of array for the `/order` endpoint?

frontend/src/basic/OrderPage/index.js
KoalaSat: this is related to what I said on the previous message.

The NaN error has nothing to do, already resolved

KoalaSat: > Ideally we do not want to supply the offset param if we have not activated the turtle mode.

Then we'll have to move up the turtle mode boolean one component up, to `OrderPage`, I was trying to avoid this, this is why I always include it

chat/views.py
KoalaSat: should I remove this? I see the response now includes it

Reckless-Satoshi: The response includes it because it is running the code in this PR :D

frontend/src/components/TradeBox/EncryptedChat/EncryptedTrutleChat/index.tsx
Reckless-Satoshi: I think it is an extra level of difficulty to have this API call if it is not needed, right? It also makes more API requests than necessary and increases latency. I might not be seeing the full picture.

frontend/src/basic/OrderPage/index.js
Reckless-Satoshi: > Then we'll have to move up the turtle mode boolean one component up, to `OrderPage`, I was trying to avoid this, this is why I always include it

Yes, it has to be at the same level as the `/api/order` call is. It might look complicated now: but it will be much cleaner once we redo TradeBox and OrderPage.

frontend/src/components/TradeBox/EncryptedChat/EncryptedTurtleChat/index.tsx
```diff
@@ -0,0 +1,304 @@
...
+    if (token && value.indexOf(token) !== -1) {
+      alert(
+        `Aye! You just sent your own robot token to your peer in chat, that's a catastrophic idea! So bad your message was blocked.`,
+      );
+      setValue('');
+    }
+    // If input string contains '#' send unencrypted and unlogged message
+    else if (value.substring(0, 1) == '#') {
+      apiClient
+        .post(`/api/chat`, {
```
Reckless-Satoshi, line 169: If we give it the offset of the last message here, the endpoint replies with the same message we sent. We can use `.then(() ...` to add it to the list of messages in addition setWaitingEcho(false) and setValue: that would make the messages sent feel way less turtle and avoid user sending the same message twice. Would feel similar to Websockets latency for sent messages.

KoalaSat: Alright I'l add it, maybe you didn't see it but `setWaitingEcho` and `setValue` are already being called on `.finally(...`

chat/serializers.py
```diff
@@ -18,7 +18,7 @@ class Meta:
 class PostMessageSerializer(serializers.ModelSerializer):
     class Meta:
         model = Message
-        fields = ("PGP_message", "order", "offset")
+        fields = ("PGP_message", "order_id", "offset")
```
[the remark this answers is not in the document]
Reckless-Satoshi: At some point (newer commit) this change has been reverted, I believe this causes on the latests commit on this branch the `"bad_request":"Order ID does not exist"` response.

frontend/src/components/TradeBox/EncryptedChat/EncryptedTurtleChat/index.tsx
Reckless-Satoshi: I saw it! it's what made me think "Oh, the spinner goes away but the new message still takes a few seconds to show on the screen"

Reckless-Satoshi:
Hey @KoalaSat , merged in 3446fc33d3c2040cc1d09881e50ab44116fe7f7f . This one was painful but now we can see the light.

Let's do 500K reward Sats for this PR too, so the whole new encrypted chat is ~1M Sat in total (similar to the Websockets reward).

Reckless-Satoshi:
> Let's do 500K reward Sats for this PR too, 

Hey @KoalaSat don't forget this one! :D

PR #315, opened 2022-11-07
Add Tradebox, OrderDetails, Notifications, OrderPage functional components.

### LINKED ISSUE #283: UX suggestion - changing title bar as additional event notification (in addition to ring)
When other party takes your order, there is a bell (audio) notification. But in Tor Browser by default NoScript disables media playback, so user will not notice that, unless manually confirming specific media file. Instead there is just this following error popup:
image

Would be cool to have some blinking / changing title as additional notificaiton option, so that user could have his Tor Browser minimized in the title bar and see notification there. You could also just be having sound on your computer muted for some reason as well. Something like https://stackoverflow.com/a/7617318/2704037.

### LINKED ISSUE #303: PRO & Basic: TradeBox and OrderPage functional components.
*This is an excerpt of main thread for RoboSats PRO* https://github.com/Reckless-Satoshi/robosats/issues/177#issuecomment-1289175371

We need to turn the current `/src/components/TradeBox` into a reusable functional component that makes use of the new `Trade` type `/src/models/trade.model.ts` (read above). This component is currently used on the `OrderPage` of RoboSats Basic, therefore this work will be useful for both frontends. 

Refactoring this component into a more reusable functional component needs of also turning the OrderPage into a functional component. The new TradeBox has to be wrapped into a simple widget for use in RoboSats Pro. These two components are pretty massive and composed themself of many little pieces that can be turned into standalone components (for example, the TradeSummary was already turned into `/src/components/TradeBox/TradeSummary.tsx`).

Turning both, the OrderPage and the TradeBox into functional components and building the Trade widget is **rewarded with 2M Sats :zap:**  drop a comment if you want to be assigned.

Reckless-Satoshi:
## What does this PR do?
Continuation of " Add TradeBox functional component #309 " . Fixes #303 . Fixes #283 
- Order Details component
- New Order Page
- TradeBox functional component and all subcomponents
- Moves every action the user takes to the Trade box (used to be disperssed: e.g. cancel and collab cancel were in the Order Box)
- New Notification Component
- Order API refresh loop now in main.tsx (notifications fire wherever the user is in the UI)

## Checklist before merging


Reckless-Satoshi:
Almost ready :eyes: Easily the largest PR so far...

Reckless-Satoshi:
It might some bugs here and there, but it is a total upgrade respect to `main` . Merging as is to keep debugging in Testnet.

PR #322, opened 2022-11-13
Move models into their own file

redphix:
## What does this PR do?
- Movies each api model into separate files
- New structure:
  ```
   api
   └── models/
       ├── order.py
       │   ├──> Order Model
       │   └──> Currency Model
       ├── payments.py
       │   ├──> LNPayment Model
       │   └──> OnchainPayment Model
       ├── market.py
       │   └──> MarketTick Model
       └── robot.py
           └──> Profile Model

  ```

## Checklist before merging


redphix:
@Reckless-Satoshi Haven't had the time to properly draft the api tasks properly. Thought I'd just begin the refactoring process with small changes for now to get the ball rolling.

redphix:
You might want to rename the imports in the "0001-inital" migrations as well. Also why are they gitignored ? I thought migrations are supposed to be commited? Not sure, haven't worked with django long enough to know if doing that is useful or even required.

redphix:
Actually I just realized, this is not a good idea. :stuck_out_tongue_closed_eyes: It's Better to make changes - even plain restructuring directory ones - in the new api altogether. You can just close this PR when you read this @Reckless-Satoshi .

Reckless-Satoshi:
> Actually I just realized, this is not a good idea. stuck_out_tongue_closed_eyes It's Better to make changes - even plain restructuring directory ones - in the new api altogether. You can just close this PR when you read this @Reckless-Satoshi .

Hey @redphix , but if you think it is better to proceed differently, let's do it like that!

Indeed, migrations should be committed. They were extremely messy at the beginning as the tables were changing back and forth every minute. But looking into more nodes running the coordinator in the future, proper migration versioning has to be contemplated.

PR #323, opened 2022-11-14
Android fixed turtle mode

KoalaSat:
## What does this PR do?
Always forces turtle mode on Android app.
Posted chat messages are displayed on the same request.

## Checklist before merging


Reckless-Satoshi:
Hey @KoalaSat . This PR looks good! Merging as is. 

Have you been able to test it? Last time I tried the turtle mode was not receiving/sending messages. I was not able to look into the Android app logs, but given that the "send" button was disabled possibly the peer_pubkey was not received.

KoalaSat:
> Hey @KoalaSat . This PR looks good! Merging as is.
> 
> Have you been able to test it? Last time I tried the turtle mode was not receiving/sending messages. I was not able to look into the Android app logs, but given that the "send" button was disabled possibly the peer_pubkey was not received.

Yes I tested it in both web and Android and it works, but truth is that I didn't test starting with an empty chat, do you think the peerpud key  problem would be related to that?

PR #324, opened 2022-11-16
Fix export chat android

KoalaSat:
## What does this PR do?
Fixes the Export chat button, when clicked on Android, it copies the content to the clipboard instead of generating a file.

## Checklist before merging


Reckless-Satoshi:
Hey @KoalaSat , this is a really neat solution for the export in Android. :rocket:

It might not be clear for the user that after clicking the export button ~nothing happens~ the log is now on the clipboard. What do you think about adding a Tooltip with "Copied!" identical to the ones we use for copying to clipboard tokens and invoices?

KoalaSat:
Yeah right, I just didn't notced because on the lastest versions of android the system notifies you what goes to the clipboard

Reckless-Satoshi:
Merging as is. I will add the tooltip and ship the full app as is in a new release.

PR #336, opened 2022-12-19
Update ca.json

BlueLynx21:
Android app phrases updated.


Reckless-Satoshi:
Very appreciated @BlueLynx21 ! We have in fact a lot of new sentences in the app since v0.3.2 that are not even in en.json. We will have to do a second round of translations for the app once we reach the v0.4.0 version :D

PR #343, opened 2023-01-31
Update telegram_watcher.py

Gravity2106:
It is used "startswith()" to check if the text start with  "/start". "filter()" instead of "get()" to avoid an exception if there is no profile asociated with a token. "attempts -= 1" instead of "attempts = attempts - 1" An "error.log" file is created if "response = self.session.get" does not work

## What does this PR do?
Fixes #<PR_NUMBER/>

## Checklist before merging
- [ ] Make sure you have installed and initialized [pre-commit](https://pre-commit.com). `pip install pre-commit` and `pre-commit install`


Reckless-Satoshi:
Hey @Gravity2106 ,

Thanks a lot for taking a look at the Telegram watcher thread.  Just a couple things before I test these fixes.
1) Can you install [pre-commit](https://pre-commit.com/) ( `pip install pre-commit` and then `pre-commit install` ) before you do "git commit". That way the formater and linter run first.
2) What about tackling this. One of the issues users have when enabling the Telegram notifications is that they go and talk to the bot "/start" without using the telegram token given by the RoboSats coordinator to their robot. Can you also add a reply to the user with instructions in case the initialization is bad? E.g. "Invalid, you must start this bot by clicking or scanning the link of the RoboSats app".

api/management/commands/telegram_watcher.py
```diff
@@ -27,40 +26,47 @@ def handle(self, *args, **options):
...
-                    except Exception:
+                if not result["message"].get("text") or not result["message"]["text"].startswith("/start"):
+                    if result["message"]["text"] == "/start":
+                        token = input("Please enter your token: ")
+                        profile = Profile.objects.filter(telegram_token=token).first()
+                        if not profile:
+                            print(f"No profile with token {token}")
+                            continue
+                if len(result["message"]["text"].split(" ")) < 2:
+                    print("Invalid format. It should be: /start <token>")
```
Reckless-Satoshi, line 51: These print statements do not go anywhere :D 

In order to send a message to the user, one has to send a message using something similar to what we have in the function `send_message()` in `api/messages.py`. That is, make a torified request to the telegram Endpoint with the `tg_api_key`, the `message` for the user and the user `chat_id` . 

We could also instead think how to make 'Telegram' class' `send_message()` a bit more generalist so we can simply call it right here for this special case. This case is special because we know the `chat_id`, but we do not know yet what Robot it is since the token was invalid :)

Reckless-Satoshi (changes requested):
Hey @Gravity2106 thanks for the fixes! I got around to try to run it. There is a couple major bugs, in addition the new functions do not work as intended. I could fix them, but since I know you are eager to learn and this fix is very low priority, do you want to have a look at my review and patch it once again? :D

Reckless-Satoshi:
Great! Added some bug fixes in a new PR over this branch https://github.com/Reckless-Satoshi/robosats/pull/354
Closing here.

PR #345, opened 2023-02-07
add e-Transfer to payment methods

toadlyBroodle:
There seems to be some misconceptions of how e-Transfers work in Canada: specifically that e-Transfers may be initiated either by the sender or the receiver. This addition is meant to add clarification for what e-Transfers are and how they work.


Reckless-Satoshi:
Hey @toadlyBroodle , thanks for enhancing the documentation for fellow Canadian robots! :rocket:

toadlyBroodle:
Absolutely, robosats is great!

PR #355, opened 2023-02-14
add Wise payment method description

toadlyBroodle:


Reckless-Satoshi:
Great! Thanks a lot for helping Robots by completing the docs for payment methods.

PR #365, opened 2023-02-24
Update ca.json

BlueLynx21:
Remaining sentences translated into Catalan.


Reckless-Satoshi:
Amazing, this is the first translation to be active and fully complete after the i18n refactorization! :rocket:

Issue #228, opened 2022-09-07
[Draft] Roadmap for RoboSats Federation


Reckless-Satoshi:
## Introduction
At the moment, anyone can spin up a new RoboSats instance. It's free source code and that's how it should be used. However spinning up new instances leads to two issues: 1) further fragmentation of an already very thin p2p market and 2) lose of revenue for further platform development (operation becomes detach from code maintenance). 

Currently there is a single RoboSats coordinator that is operated by the devs, so these two problems mentioned before are not an issue. However, this approach is also problematic as centralization makes RoboSats susceptible to attacks. In addition, given the recent attacks to Open Source devs, it would also be best if we can completely detach code development from any of the regulatory risks of operating the p2p coordinator.

In this issue I propose a solution to federate RoboSats model into many p2p coordinators competing in a market. Coordinators can each host their own frontends and, by default, show every other coordinator member orders (seamless joint order books). Users self-hosting the sovereign node app will always be able to add/delete federation members as they wish.

## Definitions
**RoboSats** the app with cool robots to exchange BTC for fiat P2P.
**Coordinator** the party in charge of allowing matchmaking and source of trust. Creates the invoices. Maintains the RoboSats backend and infrastructure. Source of trust.
**Federation** the set of all coordinators/operators of RoboSats backends.
**Devs** those contributing code.
**Dev fund** the pool of capital coming from dev tax and donations used to maintain and improve RoboSats.

#### Frontend:

- [ ] Read from `static/federation.json` the list of RoboSats coordinators and their `.onion` addresses.
- [ ] Book page requests all orders from all coordinators. Shows each row on coordinator's color as well as the coordinator logo. Book page allows for filtering by Federation member.
- [ ] If a user clicks on the order of another coordinator, it should trigger an `api/usergen/` request (so that Robot / Avatar / Token hash / PGP pubkey and encPrivKey exists as well on the coordinator's database)
- [ ] Everything else in the trade pipeline remains the same. However the `/order/` API requests will be made to that order's coordinator `.onion` address.
- [ ] Each coordinator, on its hosted frontend should be able to opt-in and out of showing other coordinators orders. They can do so by editing the static file they host in `static/federation.json`. Users who are self-hosting the frontend can delete/add as they wish.
[...]

PR #380, opened 2023-03-07
Add The Federation Layer

Reckless-Satoshi:
## What does this PR do?
This PR brings to main the experimental fork The Federation Layer. We will keep it open as a draft for some time to track progress. We will certainly keep it away from main until well tested with a bunch of Testnet voluntary coordinators.

For context read the Issue where the RoboSats Federation concept was introduced https://github.com/Reckless-Satoshi/robosats/issues/228
#### Frontend:

- [x] Read from `static/federation.json` the list of RoboSats coordinators and their `.onion` addresses.
- [x] Build Coordinator Dialog with a profile of each coordinator. Optionally, make it cool and add coordinator badges.
- [x] Build Coordinator table (keep it in the Settings tab)
- [x] Refactor Stats dialogs to show aggregated stats for all active coordinators (volume, premium, etc).
- [x] Create Context high level functions and abstractions of the federation.
- [x] Book page requests all orders from all coordinators. Shows each row on coordinator's color as well as the coordinator logo. Book page allows for filtering by Federation member.
- [x] If a user clicks on the order of another coordinator, it should trigger an `api/usergen/` request (so that Robot / Avatar / Token hash / PGP pubkey and encPrivKey exists as well on the coordinator's database). EDIT: new middleware will generate robots in the fly if auth token (SHA256(robot_token)) does not exist. New endpoint "Robot" to fetch existing robot nickname, pubkey, etc.
- [x] Everything else in the trade pipeline remains the same. However the `/order/` API requests will be made to that order's coordinator `.onion` address.
- [x] Each coordinator, on its hosted frontend should be able to opt-in and out of showing other coordinators orders. They can do so by editing the static file they host in `static/federation.json`.
- [x] Use paths for each coordinator. E.g. `...robosats.onion/order/coordinator_shortname/21` 
- [ ] If selfhosted, use the same "local"  endpoint for every coordinator, yet a different path using the coordinator shortname. E.g. ".../api/coordinator_shortname/order/...".On nodeapp, spawan a socat bridge for each Coordinator Onion address and write nginx config to proxy each API subpath to the coordinator onion endpoint.
- [ ] *Optional - Low priority* On the coordinator table, add tools for users to delete/add coordinators using a form (issue: these are forgotten in browsers that do not keep local storage after restart. This will work just fine on Android app).
- [ ] *Optional - Needs research* At the end of a trade, the user rates the coordinator (instead of "What do you think of RoboSats?"). The order ID, rating and coordinator alias is sent to every Federation member. The ratings will help keep coordinators informed about their own performance and other federation members performance. It will allow for sorting access to coordinator addresses in the docs / github / frontends according to some non-arbitrary criteria.

#### Backend:
- [x] Allow CORS. Let other origins make calls to coordinator API.
- [x] Upgrade authentication. Replace SessionID + CSRF token for only robot hash token on every authenticated request. Implement middleware that creates user+profile+avatar entries on first authenticated request with a robot token hash not seen before.
- [x] Deprecate /user/ POST.
- [ ] Anyone with the hash of a robot token could send authenticated requests (e.g, another coordinator could MITM attack a user on another coordinator). Solution: robot submits his pubkey on first authenticated requests, then every next requests must be pgp signed?
- [ ] Delete current rewards program (hard to fit on a multi-coordinator set up. There are ways, but this is to be explored in the future).
- [ ] Implement `compute_total_margin()` on each trade finished. Implement Keysend feature for Dev donations.
- [ ] *Optional - Needs research* coordinators backend could periodically fetch stats from other coordinators and "keep them in check". E.g. a coordinator might not want to show in its frontend orders of Federation members whose LN node is down, whose trade fee is very large, not contributing Sats to the dev fund, etc.

#### Other:
- [ ] Write up `FEDERATION.md` base guide/ruleset document.
- [ ] Build and release docker-compose orchestration.
- [ ] Build and release kubernetes RoboSats orchestration.


Reckless-Satoshi:
This PR is followed up in #601

PR #384, opened 2023-03-09
Capture HTTP errors on Android

### LINKED ISSUE #376: Fix torified bad requests
**Describe the bug**
The coordinator replies with Error 4XX with body `"bad_request":"reason"` in many ocassion. At the moment, these requests work without problem in the Web client. The bad request reason is used to display elements to the user. However, on the torified webview Android App these requests "fail" the try/catch in  . Therefore, there is no body response to parse, and the "reason" never makes it to the UI. 

https://github.com/Reckless-Satoshi/robosats/blob/8d4301a0c0598decab13b1b40c594c938edf36bb/mobile/services/Tor/index.ts#L33-L40

Here the response body is not returned in the `reject()` in act, `torDaemon` seem to not be fit to return the body if status is 400.

https://github.com/Reckless-Satoshi/robosats/blob/8d4301a0c0598decab13b1b40c594c938edf36bb/mobile/App.tsx#L61-L67

In addition, here the `error` is "lost" since the `.catch()` returns `sendTorStatus` instead of `error`.

**To Reproduce**
This issue can be observed everytime the coordinator replies with a 400 status. Example:
1) Recover a robot with token the token `aaaaaaaaaaaaaa`. You expect the coordinator to return with a 400, reason "Token does not have enough entropy", however, on the torified app the request to create a new robot never seems to finish.
2) Fetch an order book that is empty. The response is 400 "No orders". However the Android app seems to never finish fetching the book. Can be tested using the Testnet  switch (as testnet usually never has active orders).
3) Create an order then cancel it. The coordinator replies with a 400 "Maker cancelled the order" when you cannot access the order anymore. However it seems 

**Expected behavior**
The android torified app should behave the same way as the Web Client.

**Additional context**
I have tried 3 different quick ways to solve this issue none of which work. Can be seen in branch https://github.com/Reckless-Satoshi/robosats/tree/fix-torified-bad-requests .

Another workaround is to never send a response that has a useful body with `status.HTTP_400_BAD_REQUEST` from the coordinator.

KoalaSat:
## What does this PR do?
Fixes #376

This PR makes react-native to capture and return error responses to the JS app

## Checklist before merging
- [ ] If I added new phrases to the user interface, I have ran prettier `cd frontend/static/locales; python collect_phrases.py` to collect them for translation.

### DIFF
--- a/mobile/App.tsx
+++ b/mobile/App.tsx
@@ -54,16 +54,18 @@ const App = () => {
   };
 
   const onCatch = (dataId: string, event: any) => {
-    let json = '{}'
+    let json = '{}';
     if (event.message) {
-      const reponse = /Request Response Code \((?<code>\d*)\)\: (?<json>\{.*\})/.exec(event.message)
-      json = reponse?.groups?.json ?? '{}'
+      const reponse = /Request Response Code \((?<code>\d*)\)\: (?<json>\{.*\})/.exec(
+        event.message,
+      );
+      json = reponse?.groups?.json ?? '{}';
     }
     injectMessageResolve(dataId, {
       headers: {},
-      json: JSON.parse(json)
+      json: JSON.parse(json),
     });
-  }
+  };
 
   const onMessage = async (event: WebViewMessageEvent) => {
     const data = JSON.parse(event.nativeEvent.data);
--- a/mobile/App.tsx
+++ b/mobile/App.tsx
@@ -53,6 +53,18 @@ const App = () => {
     loadCookie('garage').then(() => injectMessageResolve(reponseId));
   };
 
+  const onCatch = (dataId: string, event: any) => {
+    let json = '{}'
+    if (event.message) {
+      const reponse = /Request Response Code \((?<code>\d*)\)\: (?<json>\{.*\})/.exec(event.message)
+      json = reponse?.groups?.json ?? '{}'
+    }
+    injectMessageResolve(dataId, {
+      headers: {},
+      json: JSON.parse(json)
+    });
+  }
+
   const onMessage = async (event: WebViewMessageEvent) => {
     const data = JSON.parse(event.nativeEvent.data);
     if (data.category === 'http') {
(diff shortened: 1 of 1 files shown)


Reckless-Satoshi:
Hey @KoalaSat 

Thank you for this PR! It certainly fixes a lot of small broken flows in the Android App. Tested and works as expected for the little ride I gave. Can be merged as is, high quality PR! :rocket:

Please drop an invoice from a proxy wallet with a long expiration time for 400K Sats :zap:

KoalaSat:
lnbc4m1pjq5cxmpp5jfwy9e2rp2zj4jkv7jxxsfzyyd39t998f49juqqmdj6lyu8mvr7qdqu2askcmr9wssx7e3q2dshgmmndp5scqzpgxqyz5vqsp5e4v2un69nptk0wl7xwylm8wrwpm0hjzyyjcmzvy7q9pj92ajzuks9qyyssq6l77dcctpn60qjgzn47mjackzns63e76zwxjzxa3qdr3rya4p6l36qhzs6vfe4dshwhnpr6qtlxhew4qp5p70spuhj2q0wjl9q3aadqq9hquhs

Reckless-Satoshi:
> lnbc4m1pjq5cxmpp5jfwy9e2rp2zj4jkv7jxxsfzyyd39t998f49juqqmdj6lyu8mvr7qdqu2askcmr9wssx7e3q2dshgmmndp5scqzpgxqyz5vqsp5e4v2un69nptk0wl7xwylm8wrwpm0hjzyyjcmzvy7q9pj92ajzuks9qyyssq6l77dcctpn60qjgzn47mjackzns63e76zwxjzxa3qdr3rya4p6l36qhzs6vfe4dshwhnpr6qtlxhew4qp5p70spuhj2q0wjl9q3aadqq9hquhs

1d1fb557e576666cec1816ac2fce24ea55db353bdee234fb759ca94a344e6015

PR #392, opened 2023-03-16
Added Serbian Dinar (RSD) currency.

lukabb:
Verified that the currency is present in: https://api.yadio.io/exrates/BTC


Reckless-Satoshi:
Looks good! :rocket:

PR #393, opened 2023-03-17
Added L-USDt as payment method

minaraus:
## What does this PR do?
This PR add L-USDt as payment method.


minaraus:
index.tsx changed a lot. The code will be harder to review.
Another PR will be opened instead.

PR #406, opened 2023-03-30
Fixed npub address

jimtoshi:
Had the npub for the Robosats order bot in there previously. Changed to the official Robosats npub.


Reckless-Satoshi:
Looks good!

PR #412, opened 2023-04-09
Added Iranian Toman (IRT) currency

HT696:
Added Iranian Toman (IRT) currency


Reckless-Satoshi:
Hey @HT696 , looks good! Thank you for this contribution.

Merging as is, however IRT will only work after my next commit to main patching the overflow Error of the `DecimalField` due to the insanely cheap price of IRT.

Issue #266, opened 2022-10-03
Refactor API Design (API v1)


redphix:
**Status: Draft**

This issue aims to document and outline a new refactored API design for Robosats (API v1) with some improvements and discussions about changes.

### Routes Overview

- **`/robot`**
  - `POST /robot` - Get me a new robot (conventional `/login` route)
  - `GET /robot`  - Get your robot info (conventional `/profile` route)
  - `PATCH /robot/preferences` - Update your preferences (things like stealth invoices, tg enabled/disable)
  - `POST /robot/withdraw-reward` - Claim/Withdraw rewards
- **`/book`**
  - `GET /book`
- **`/exchange`**
  - `GET /exchange/info`
  - `GET /exchange/history`
  - `GET /exchange/limits`
- **`/market`**
  - `GET /market/prices`
  - `GET /market/ticks`
- **`/order`**
  - `POST /order` - Create a new order
  - `GET  /order/:id`
  - `POST /order/:id/take`
  - `POST /order/:id/pause`
  - `POST /order/:id/invoice`
  - `POST /order/:id/address`
  - `POST /order/:id/cancel`
  - `POST /order/:id/confirm`
  - `POST /order/:id/dispute`

#### Notes - Routes
- Having an hierarchy as described above, we can also get a nice hierarchy in our
  codebase. Each of /market, /order, /exchange, /book and /robot can be a
  django ViewSet and under each viewset we can have separate views in separate
  .py files to organize the code better.
- If you create an order you are implicitly a maker, so don't really required
  to mention it as `/make` or `/order/make`, `POST /order` conveys the
  intention of "making a new order"
- skip `submit_statement` action and add as part of `/dispute` route itself

### Things to improve from old design
1. Define each route to return only one type of response type (i.e the response
  model should be deterministic). In other words each route's json response
  body's keys must always be deterministic, no matter the input. Try using
  nested model/serializers if it's not possible to fit all types of response in
  a single model. At most, as a last resort, you may define another response
  model that may be returned.

2. Use arrays where ever sequential data is involved as the order matters.<br>
   Currently, the /historical route returns a response like this:
   ```json
   {
       "2022-02-22 00:00:00+00:00": {
           "volume": 0.001,
           "num_contracts": 3
       },
       "2022-02-23 00:00:00+00:00": {
           "volume": 0.004,
           "num_contracts": 2
       },
       "2022-02-24 00:00:00+00:00": {
           "volume": 0.003,
[...]

PR #481, opened 2023-04-30
Refactor models into a module

Reckless-Satoshi:
## What does this PR do?
First tiny step towards API v1 (https://github.com/Reckless-Satoshi/robosats/issues/266). Models have been refactored into a model where each model is its own file. Circular dependencies have been solved.

The Profile model has been renamed as Robot.


redphix:
:clap:

PR #609, opened 2023-05-22
Additional updates to LnProxy support

shyfire131:
## What does this PR do?
1. Fixes the failed action issue raised in matrix devchat
2. Corrects where the relay list is being fetched from - don't fetch from fork anymore
3. Bumps the `action/checkout` dependency from v2 to v3

Regarding the reason that the action was failing, this was most likely because the Robosats workflow permissions are set to the most restrictive setting:

---
image
---

More info on the permissions syntax used can be found here:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

## Checklist before merging
- [N/A] If I added new phrases to the user interface, I have ran `cd frontend/static/locales; python collect_phrases.py` to collect them for translation.

### DIFF
--- a/.github/workflows/lnproxy-sync.yml
+++ b/.github/workflows/lnproxy-sync.yml
@@ -3,11 +3,14 @@ name: Sync lnproxy relays
 on:
   schedule:
     # * is a special character in YAML so you have to quote this string
-    - cron:  '0 12 * * 0'  # Run every Sunday at noon
+    - cron: "0 12 * * 0" # Run every Sunday at noon
 
 jobs:
   sync:
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
 
     steps:
       - name: Checkout
--- a/.github/workflows/lnproxy-sync.yml
+++ b/.github/workflows/lnproxy-sync.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Fetch and process JSON
         run: |
-          curl https://raw.githubusercontent.com/shyfire131/lnproxy-webui2/17d6131a72a92978e5c0dc57ab2b2fbe467c7722/assets/relays.json -o lnproxy_tmplist.json
+          curl https://raw.githubusercontent.com/lnproxy/lnproxy-webui2/main/assets/relays.json -o lnproxy_tmplist.json
           node .github/workflows/scripts/lnproxy-sync.js
 
       - name: Commit and push if it's not up to date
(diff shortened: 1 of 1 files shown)


Reckless-Satoshi:
Thank you for this update! Merging it :rocket:

I am unsure whether relaxing the permission settings for Actions is the way to go. Mostly because we make heavy use of actions we do not directly mantain, probably this won't be an issue whatsoever but better safe than sorry ^^

Do you think it is also possible to make this workflow open a PR once it detects a new lnproxy relay? Similarly to how DependaBot opens PRs

shyfire131:
> Do you think it is also possible to make this workflow open a PR once it detects a new lnproxy relay? Similarly to how DependaBot opens PRs

This is possible, however we would still need to keep the `contents: write` permission since the workflow needs to create a branch in order to create a PR.

I can't see any way around this unfortunately (doesn't mean there isn't one, just that I can't see one).

One option is to enable branch protection rules for `main` in the robosats repo. That would effectively result in the same permissions: workflows can't write to main, but they can create branches and submit PRs. 

The effective permissions for Actions would then look as follows:

---
image
---

Let me know how you'd like to proceed - I have the updated workflow that creates a PR ready to roll

Reckless-Satoshi:
I am not sure I understand the screenshot correctly. We currently have this setting as "Read repository contents and packages permissions". Should it be changed to "Read and write permissions" ? Currently this repository Actions settings match this screenshot, except for eh "Allow Github Actions to create and approve pull requests", this field is greyed out.
screen


Maybe worth mentioning we had to add this line to the release workflow for it to now error out of lack of permissions: https://github.com/RoboSats/robosats/blob/5ff70bccb75bde4da5fccbfd83a5f8645f4c2364/.github/workflows/release.yml#LL9C1-L10C23 Maybe this line is also needed in this workflow for it to have access to the `GITHUB_TOKEN` ?

In any case, I added protection rules to the main branch. So we can enable the permissions (if needed) for the workflow to create Pull Requests :ok_hand:

shyfire131:
> Should it be changed to "Read and write permissions" ?

No. The only thing that needs to change is that the greyed out checkbox must be ticked so that the workflow can create the PR. I'm not sure why it's greyed out, but it could be because of this:

https://stackoverflow.com/questions/72376229/github-actions-is-not-permitted-to-create-or-approve-pull-requests-createpullre/73155199#73155199

Could you take a look?

I took a look at at the release workflow. We will not need `write-all` for this workflow, of that much I am certain.

OK cool, I'll submit another PR that changes the workflow from directly committing on main to rather create a branch and submit PR.

PR #611, opened 2023-05-22
Add core-lightning as backend lightning node vendor

### LINKED ISSUE #196: Add core-lightning as LN node provider.
We are currently relying on [LND](https://github.com/lightningnetwork/lnd) for RoboSats backend node. It does work pretty well. There is only a few things that might be critical: channel.db is hard to manage and performance/reliability could still be better (I guess?). 

It would be great to implement [core-lightning](https://github.com/ElementsProject/lightning) as backend, not as a replacement, but for exploration and extra redundancy. The higher performance and use of Postgres as database would be much appreciated. While it is not clear yet whether there will be some shenanigan or unexpected issue, the redundancy of having two node providers will always be of great value.

If core-lightning is available as node provider, future effort to improve RoboSats availability could be put into creating a "LN load balancer" that will smartly direct invoice creation / payments to a pool of nodes of different providers.

We need a new [`api/lightning/node.py`](https://github.com/Reckless-Satoshi/robosats/blob/main/api/lightning/node.py) (LNnode provider) that can perform all of the same functions using core-lightning as backend instead of LND. 

- [ ] **Prerequisites not met** By @louneskmt we need a hodl invoice plugin in Core-Lightning. Out of scope for RoboSats.
- [ ] The first step would be to add a [core-lightning container](https://hub.docker.com/r/elementsproject/lightningd/tags) running testnet3 to our dev [docker-compose](https://github.com/Reckless-Satoshi/robosats/blob/main/docker-compose.yml) orchestration. 
- [ ] It might be useful to implement unit tests `test.node.py` to learn about the current set up. The unit tests will be very helpful in the future, as they should receive exactly the same behavior regardless of whether the call was attended by LND or core-lightning  in any future release of these.
- [ ] The bigger chunk of the work: write all core-lightning calls for every function.

I am unsure how big this task really is. But it certainly requires some very specific knowledge on interacting with Lightning node APIs.

**Additional context**
This task is ⚡**rewarded with 2M Sats**⚡ . Drop a comment if you want to be assigned.

Reckless-Satoshi:
## What does this PR do?
Fixes #196 . This PR adds support for core-lightning (CLN) as lightning node vendor. This PR merges into main the work of @daywalker90 for the rewarded task #196 .

Coming from : 
https://github.com/RoboSats/robosats/pull/593
https://github.com/RoboSats/robosats/pull/435
https://github.com/RoboSats/robosats/pull/433
https://github.com/RoboSats/robosats/pull/431

**Some considerations**
Using CLN on mainnnet as the RoboSats backend is strongly discouraged. Yet, using it on testnet and breaking things, is strongly encouraged. CLN has been very little tested as a RoboSats backend compared to LND, and it is likely we still have to find and fix some bugs.

Only the Dockerfile in `docker/cln` will have all the necessary libraries to work as a RoboSats backend. Our Dockerfile currently builds CLN form `ElementsProject/lightning` and the grpc plugin from `daywalker90/lightning#hodlvoice` GitHub repositories . Any other docker image will not work.

Quoting @daywalker90 in #593
> If an invoice is close to expiry (60s before) the htlcs will all be returned to the sender and the invoice will have the Hodlstate Canceled and the regular invoice lookup with cln will be UNPAID and shortly after EXPIRED.
This is probably also the biggest difference in behaviour to the lnd hodl-invoices, which is why we had to put a big expiry on all hodl-invoices in robosats.

## Checklist before merging


Reckless-Satoshi:
Hey @daywalker90 , It seems we have 2 undefined `failure_reason` in line 470 and 475 . I have been checking but not sure how to assign them the right value. These are supposed to be strings from those in `payment_failure_context` . What would be the appropriate fix here?

daywalker90:
btw in practice i've never seen it return with a non-error "pending" i wouldn't even know when this can happen

Reckless-Satoshi:
> btw in practice i've never seen it return with a non-error "pending" i wouldn't even know when this can happen

In lnd.py I see these `if` conditions just have a `pass`. Of course, it keeps cycling in the for loop until a `return` is reached. 

I added a commit defining these two `failure_reasons` with their closest match. These will be directly returned to the client with the response to a request to claim rewards from the robot profile. If this sounds good to you, I will proceed and merge into main.

daywalker90:
yes looks good.

PR #614, opened 2023-05-23
Fix bug in discount field

### LINKED ISSUE #532: It's difficult to enter a discount because the field forces a zero into it
The premium field somehow forces a 0 into itself. So when you type 1, it will type 01, you can't backspace the 0. Typing a `-` will hence be ignored because `0-` doesn't make sense. So the only way to input a discount is to use the left arrow, type `-` to get `-0`, then if you want 5% discount you type `5` to get `-50` and then you remove the `0` in the end.

Pretty annoying. Can the premium field just be a regular field without the 0 being forced into it?

JooVLC:
Fixing the type handlePremiumChange function to handle the discount to the right type (React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement>)

Using the correct type of handlePremiumChange function, then deconstructing the "value" var from the event.target

Using Number(value) in declaration of "newPremium", instead of using multiplication with number and string

To not affect another methods using "premium" var, I used isNaN func to make a check and set the initial value of "premium" var to zero (if is NaN) or newPremium (when valid number is entered)

The fix in the bug is the line "premium: isNaN(newPremium) || value === '' ? '' : premium," in setMaker. That makes the "premium" var to be set to zero when the input is empty or not a number, allowing the user to input the "-" (minus sign) and not entering 0 as before.

## What does this PR do?
Fixes #532

This PR fix the bug in discount field changing the function `handlePremiumChange` in "frontend/src/components/MakerForm/MakerForm.tsx". 
This changes "frontend/static/locales/collect_phrases.py" too, because the pre-commit was running in loop, changing the locales's json. I added a comparison in the python script to not update the json if he no has changes.

## Checklist before merging

### DIFF
--- a/frontend/src/components/MakerForm/MakerForm.tsx
+++ b/frontend/src/components/MakerForm/MakerForm.tsx
@@ -181,27 +181,28 @@ const MakerForm = ({
     });
   };
 
-  const handlePremiumChange = function (e: object) {
-    const max = fav.mode === 'fiat' ? 999 : 99;
-    const min = -100;
-    const newPremium = Math.floor(e.target.value * Math.pow(10, 2)) / Math.pow(10, 2);
-    let premium: number = newPremium;
-    let badPremiumText: string = '';
-    if (newPremium > 999) {
-      badPremiumText = t('Must be less than {{max}}%', { max });
-      premium = 999;
-    } else if (newPremium <= -100) {
-      badPremiumText = t('Must be more than {{min}}%', { min });
-      premium = -99.99;
-    }
-    updateCurrentPrice(limits.list, fav.currency, premium);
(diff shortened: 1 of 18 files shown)


Reckless-Satoshi (approved):

Reckless-Satoshi:
If you agree, I believe this PR is ready to be marked as ready for review (and subsequently merged :rocket: )

Please, accept a tip of 40K Sats from the devfund. Submit an invoice with a long expiration time from a proxy node id :)

JooVLC:
> If you agree, I believe this PR is ready to be marked as ready for review (and subsequently merged :rocket: )
> 
> Please, accept a tip of 40K Sats from the devfund. Submit an invoice with a long expiration time from a proxy node id :)

I agree, I will change the PR.

PR #647, opened 2023-06-05
Fix typo: if you if you -> if you

dergigi:
## What does this PR do?

Fixes a small word-repetition typo by replacing instances of `if you if you` with a single `if you`.


Reckless-Satoshi:
Good catch! Thank you for going through the pain of updating the locales too :rocket:

PR #673, opened 2023-06-21
Updated BBW to Blink Wallet

f321x:
Updated the Wallet Compatibility list. Bitcoin Beach Wallet is now Blink Wallet. Hodl invoices don't work correctly in Blink Wallet yet.

Couldn't get the hyperlink to the paragraph to make the Blink in the table clickable, this should be added before merging.


Reckless-Satoshi (commented):
Thank you for taking the time to update the wallets docs and let us know  that hodl invoices do actually get charged. :rocket:

I added a commit to fix the paragaph link.

@openoms  any idea why Blink is sometimes charging the hodl invoices? Before merging I'd like to be sure Blink is certainly having issues with hodl invoices. I had the impression BBeach/Blink was playing very well with Robo.

Reckless-Satoshi:
Hey @f321x , thank you for the extra context! We are going to merge as is.

> 2\. @nicolasburtey says it's better to not use Blink to pay hodl invoices currently

I wonder about this. Every invoice is potentially a hodl invoice, as the sender cannot know how long a payment will take to resolve. So I guess this is a pressing issue to solve in Blink. Would be awesome to get notified once long-lived payments are handled correctly so we edit the Blink docs once again.

nicolasburtey:
the hodl invoice issue has been fixed already - and all previous funding from hold invoice that has been cancelled has been returned to their users @Reckless-Satoshi @f321x

f321x:
Tested the new, fixed version, hodl invoices work fine again so i created a new PR with updated description at #686

PR #686, opened 2023-06-25
Update Blink wallet in wallet compatibility documentation

f321x:
Updated Blink description, bond and conclusion symbol in table, and version number in the wallet compatibility docs as hodl invoices now work fine and the wallet can be used with Robosats without limitations.
Related to PR #673


Reckless-Satoshi (approved):
Looks good, merging! 
Thank you for taking the time to track the compatibility with Blink and report for the rest of the robots :rocket:

PR #692, opened 2023-06-30
Update handleCurrencyChange to not reset amount range on currency change in maker form

f321x:
Currently when changing the currency in the advanced maker form the previously set amount range is discarded and reset to the standard values. This is pretty annoying when the user sets the amount range first and then changes the currency. 
This PR changes the handleCurrencyChange function to only reset the amount range if the current values are out of range of the newly selected currency and not unnecessarily reset it on every currency change.

Tested with the local docker env, seemed to work as supposed.

### DIFF
--- a/frontend/src/components/MakerForm/MakerForm.tsx
+++ b/frontend/src/components/MakerForm/MakerForm.tsx
@@ -140,11 +140,17 @@ const MakerForm = ({
     updateCurrentPrice(limits.list, newCurrency, maker.premium);
 
     if (makerHasAmountRange) {
-      setMaker({
-        ...maker,
-        minAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.25).toPrecision(2)),
-        maxAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.75).toPrecision(2)),
-      });
+      const min_amount = parseFloat(Number(limits.list[newCurrency].min_amount).toPrecision(2));
+      const max_amount = parseFloat(Number(limits.list[newCurrency].max_amount).toPrecision(2));
+      if (parseFloat(maker.minAmount) < min_amount || parseFloat(maker.minAmount) > max_amount
+      || parseFloat(maker.maxAmount) > max_amount || parseFloat(maker.maxAmount) < min_amount)
+      {
+        setMaker({
+          ...maker,
+          minAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.25).toPrecision(2)),
+          maxAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.75).toPrecision(2)),
+        });
+      };
     }
   };
 
--- a/frontend/src/components/MakerForm/MakerForm.tsx
+++ b/frontend/src/components/MakerForm/MakerForm.tsx
@@ -140,15 +140,15 @@ const MakerForm = ({
     updateCurrentPrice(limits.list, newCurrency, maker.premium);
 
     if (makerHasAmountRange) {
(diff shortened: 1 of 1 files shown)


Reckless-Satoshi:
Hey @f321x thank you for this PR! Indeed, the handling of ranges and currency changes leads to occasional  frustrating experience. I will be giving it a test in the following days, I'm unfortunately very busy lately with fiat life :`)

frontend/src/components/MakerForm/MakerForm.tsx
```diff
@@ -140,11 +140,17 @@ const MakerForm = ({
...
-        maxAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.75).toPrecision(2)),
-      });
+      const min_amount = parseFloat(Number(limits.list[newCurrency].min_amount).toPrecision(2));
+      const max_amount = parseFloat(Number(limits.list[newCurrency].max_amount).toPrecision(2));
+      if (parseFloat(maker.minAmount) < min_amount || parseFloat(maker.minAmount) > max_amount
+      || parseFloat(maker.maxAmount) > max_amount || parseFloat(maker.maxAmount) < min_amount)
+      {
+        setMaker({
+          ...maker,
+          minAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.25).toPrecision(2)),
```
Reckless-Satoshi, line 150: We could be using here `min_amount` and `max_amount` given that we defined them already in L143 and L144.

frontend/src/components/MakerForm/MakerForm.tsx
```diff
@@ -140,11 +140,17 @@ const MakerForm = ({
     updateCurrentPrice(limits.list, newCurrency, maker.premium);
 
     if (makerHasAmountRange) {
-      setMaker({
-        ...maker,
-        minAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.25).toPrecision(2)),
-        maxAmount: parseFloat(Number(limits.list[newCurrency].max_amount * 0.75).toPrecision(2)),
-      });
+      const min_amount = parseFloat(Number(limits.list[newCurrency].min_amount).toPrecision(2));
```
Reckless-Satoshi, line 143: It seems to be the agreed JS style to use camelCase to name these consts, so probably best to rename them to `minAmount` and `maxAmount`

Reckless-Satoshi (changes requested):
Looks good! I drop two small comments in case you want to add a new commit fixing these :D Then it will be ready to merge.

f321x:
Updated the Functions to respect camelCase and use the existing constants. Thx for having a look.

Reckless-Satoshi (approved):
Looks good! Tested and works as expected. :rocket: Please accept a small tip of 20K Sats, submit a LN invoice from a proxy node id with a long expiration time (+24h if possible).

The whole advanced options are still a bit convoluted, so we will have to rework them in the future. Good enough for now :D

Reckless-Satoshi:
> lnbc200039500p1pj2hm6wpp5vzdsaq7zhv0lltpr8f5x3hehsqus5xxsvvlgstzwxt8e0ge63dhsdqjwfhkymmnv968xgzs2gcqza3xqyjw2wsp5rulatued7palweke0rv0jkh4mstf5xx255vk3verc82su0swmhaq9qyyssquufmvxqgjp7l98lqtm6wpkum5xgp4nz0dcx7yh7gf6wq3rk9xu3834y8e63wyeq7rng35pf4axmggv3x6l7n066eemf44pl3y685hdgplt7m9z

4f994d2b90ca53201e7a87e27b7f113a1b6d45cbe3e68bde8efe1312e3e00e5a

PR #764, opened 2023-08-03
Add coordinator panel order logs

Reckless-Satoshi:
## What does this PR do?
This feature is intended to help new RoboSats coordinator understand and debug the pipeline of the active orders, effectively, making them more capable to provide user support if needed.

This PR ads a new `TextField` to the `Order` model: `logs` . All relevant events of the order lifecycle are logged and rendered with HTML formatting on the admin panel.

## Checklist before merging


Reckless-Satoshi:
Looking forward to improve a bit more over time the logging for the orders. At the moment this is sufficient to be _very_ useful for new coordinators. The logs are formatted as an html table and hyperlinks to the objects mentioned are inserted during the admin panel rendering for easier navigation.
image

PR #801, opened 2023-08-20
fix typo

douglaz:
This creates a broken link on the site


Reckless-Satoshi (approved):
Hey @douglaz , thank you for finding it out! This typo is also fixed by #802 . I will however, merge your first contribution and then solve the conflict with Okada's PR, the more contributors the merrier! :D

PR #807, opened 2023-08-23
Ditch react-native-tor in favor of official Tor Android client

shyfire131:
## What does this PR do?
This PR lays foundations for https://github.com/RoboSats/robosats/issues/619

This PR removes the (unmaintained) dependency react-native-tor, and replaces it with the [official Tor android proxy](https://github.com/guardianproject/tor-android/tree/d30c3e73dbff79bc99f095028da064618b4dcdd6). It wraps the official Tor android proxy in a React Native [Native Module](https://reactnative.dev/docs/native-modules-android) and starts it as a foreground service on app boot. 
The TorClient class in React Native is updated to use OkHttp, which is configured to use the new Tor Proxy.

Bonus: since we're improving Tor, all instances of "TOR" have been renamed to "Tor" to align with the official spelling.

Roadmap items for future consideration
- [ ] Use React Turbo Modules (New Architecture) 
- [ ] If a Tor proxy is already running e.g. Orbot, there is no option to configure RoboSats to use that. It will start a second proxy. I think this is mostly fine though. 
- [ ] The app needs a logging module, ideally something where logs can be exported so that users can submit diagnostic logs when they encounter issues
- [ ] Websockets :)

## Checklist before merging
- [ ] smoke testing: Test a trade end-to-end, currently the chat has not been tested since the site is down
- [ ] some soak testing would be good too, big change
- [ ] Proguard is currently disabled


Reckless-Satoshi:
Hey @shyfire131 this PR is looking super exciting. Please, let me know if I can help by testing out stuff :D

Let's insta-convert this PR into a rewarded task of the same value as the original torification implemented by @KoalaSat of 2M Sats

Issue #760, opened 2023-07-29
Add location to F2F payment option


f321x:
**Is your feature request related to a problem? Please describe.**
Currently, if i understand correctly, the F2F payment option has no possibility to set a location (eg City or coordinates). This renders the otherwise private and useful payment method fully useless as the trade partner cannot see where the other side is located before accepting the order and paying the bond. In Bisq for example you can see the location of the p2p trade before accepting the offer, so you can only accept it if the location is reachable/useful. This is probably also the reason why there are so few F2F offers active.

**Describe the solution you'd like**
Add a field to enter the location when creating an offer with F2F payment method and show the location in the orderbook either always (maybe if only F2F is accepted) or for example on hover over the F2F icon or when looking at the offer details.

**Additional context**
For example see Bisq where this is being done for a long time.
![Bildschirmfoto vom 2023-07-29 23-27-18](https://github.com/RoboSats/robosats/assets/51097237/b2eecd2b-a332-4cb6-a305-2cd7cc810706)

Reckless-Satoshi:
Hey @f321x thank you for bringing this up.

## How it can be done now
The payment methods field is a free text entry. The autocomplete box a faster way to enter your freestyle text, then the payment methods icons are simply parsed from the string. One could write "Face to face Amsterdam" as a payment method. Or even select "F2F" to get the icon and then enter "Amsterdam"
![image](https://github.com/RoboSats/robosats/assets/90936742/24c7e72a-7392-4331-a8ed-69e773243bb0)
![image](https://github.com/RoboSats/robosats/assets/90936742/f56bb807-29a8-4d30-97d6-4db2dd67c6db)
However, this is certainly... more of a hack than a feature :) It is not obvious for any user that the F2F method can be used this way (but I've certainly seen this a few times in the order book!). I think we should take F2F seriously and make it more clear.

## How it can be improved
I was exploring ideas some time ago, but I never got to write them down or implement them. It would be great if:
1. Selecting F2F opens a small pop up dialog with a tiled map.
2. The user clicks on the map.
3. The frontend adds a random jitter of about  10 Km (lat and lon) to make sure to anonymize a users who might click on top their own house (lol)
4. The order book shows a map with the location of all the F2F orders posted.

This will only require the `Order` model to have 2 new optional fields: `latitude` and `longitude`. I think there might be ways we can even embed a very low resolution tiled map within the client. That way we do not include any dependency to third party APIs for positioning.

[...]

PR #867, opened 2023-09-30
Add location to F2F payment option (backend)

KoalaSat:
## What does this PR do?
Backend work for #760

This PR introduces the new `latitude` and `longitude` attributes for an Order so users can assign a geographical coordinate for F2F payments.

This only includes the back-end changes.

## Checklist before merging


api/migrations/0043_order_latitude_order_longitude_alter_order_bond_size.py
```diff
@@ -0,0 +1,52 @@
...
+            field=models.DecimalField(
+                decimal_places=2,
+                default=3.0,
+                max_digits=4,
+                validators=[
+                    django.core.validators.MinValueValidator(1.0),
+                    django.core.validators.MaxValueValidator(15.0),
+                ],
+            ),
+        ),
```
KoalaSat, line 51: This was generated together with `makemigrations`, not sure if it was just missing or a mistake on my side

docker-compose.yml
```diff
@@ -61,7 +60,7 @@ services:
       - ./frontend/static:/usr/src/robosats/static
 
   clean-orders:
-    image: backend
+    build: .
```
KoalaSat, line 63: I include this not as change but more like a question, docker-compose was not able to find any `backend` since it's not defined anywhere. This was the only way I manage to make it run, am I missing something?

api/migrations/0043_order_latitude_order_longitude_alter_order_bond_size.py
Reckless-Satoshi: It is not a mistake, but it is an undesired change. It is best to delete lines L39 to L51 before merging (that is `migrations.AlterField(...)` . Alternatively, you can also delete the migration file, edit `.env` to `MIN_BOND_SIZE = 2` (instead of current value of 1) and the run `manage.py makemigrations` again. Should yield the same migration without altering the field `bond_size`.

The reason why this happened is that `/api/models/order.py` Order field `bond_size` depends on two config parameters that are read from the `.env` file. In my setup the `.env` file defines the `MIN_BOND_SIZE` as 2%, while the `.env-sample` defines it as 1% (at some point, there was a deviation, I changed my own `.env` without updating the `.env-sample`...). This migration will therefore allow bond sizes to be smaller than the current minimum of 2%.

I have opened a new issue (https://github.com/RoboSats/robosats/issues/873) so we make sure backend models are not dependant on any coordinator specific setting (`.env`) and therefore, we do not have any sort of conflict at run time when coordinators change values on their `.env` file.

Reckless-Satoshi (changes requested):
Hey @KoalaSat the PR looks good to go for me! It is best to delete the `bond_size` alter. 

I left a comment with a deeper explanation of what has gone wrong and opened an issue so we do not have this problem during future development again.

api/migrations/0043_order_latitude_order_longitude_alter_order_bond_size.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 4.2.5 on 2023-10-03 20:12
+
```
Reckless-Satoshi, line 2: Sorry to bother again :D It would be best, in order to keep migration history clean, to rename this migration from `0043_order_latitude_order_longitude_alter_order_bond_size.py` to `0043_order_latitude_order_longitude.py`

Reckless-Satoshi (approved):

Reckless-Satoshi:
This PR is ready for merge!

This PR is part of the "face-to-face" feature that has a development reward, I think it's probably best to payout the reward fully once the frontend work PR is ready. Talking about this reward, it now looks too small for the amount of work needed to get it done nicely following the current plan, so I propose doubling the reward to 2M Sats.

PR #902, opened 2023-10-19
Update dead link for developer rewards

Purchase73:
## What does this PR do?
Fixes dead link for developer rewards, in readme.md.

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Alice wants to buy satoshis privately:
 Check out our [Contribution Guide](https://learn.robosats.com/contribute/) to find how you can make RoboSats great.
 
 ### ⚡Developer Rewards ⚡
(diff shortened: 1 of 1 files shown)


Purchase73:
My first PR here, just dipping my toe in, please let me know anything I should fix. I assume the items in CONTRIBUTING.md don't apply, since this is a doc change.

Reckless-Satoshi:
Hey @Purchase73 thank you for catching and fixing this issue! Meging as is :rocket:

Moving the Developer Rewards panel into the RoboSats organization is a long overdue task. If we keep the Developer Rewards panel under @Reckless-Satoshi user, linking new issues/tasks is very inconvenient. Unfortunately, GitHub does not have a "Transfer Project" feature. So it seems that manually recreating the current project item by item in the organization is the only way forward: a total bummer and the reason we have not updated the rewards panel for a few months now.

PR #912, opened 2023-10-24
Update Amazon GC Payment Method

TimothyShimmin:
Amazon no longer accepts payment by gift card to buy gift cards. Added a potential alternative instead but there are more privacy concerns when buying from a user's wishlist, and vice versa. Addresses are typically censored but Amazon defaults to sending a note inside the package with the buyer's name, for example. 

Purse.io does something similar, might want to look into their FAQ for inspiration.

## What does this PR do?
Fixes #<ISSUE_NUMBER>

## Checklist before merging


Reckless-Satoshi:
Hey @TimothyShimmin thank you very much for updating the docs regarding Amazon GC :rocket:

PR #948, opened 2023-11-15
Fix Order page

KoalaSat:
## What does this PR do?
Refactors and remove the use of `CurrentOrder`in `FederationContext`in behalf of the order associated to the current Slot in the garage.

- Manage the discord between different coordinator fetching the same robot but with different order history
- There is no a global `CurrentOrder` anymore
- Order page is now accessible from Robot page, NavBar and the orders Book.
- Order page 100% relays on the params information `exp/12345` to fetch and display order data
- Refactor and removed some unnecessary calls to some Contexts

⚠️ This PR only fixes the access to the Order page and the initial display, I didn't check yet the order workflow.

## Checklist before merging


frontend/src/models/Garage.model.ts
```diff
@@ -3,9 +3,22 @@ import { systemClient } from '../services/System';
 import { saveAsJson } from '../utils';
 export interface Slot {
   robot: Robot;
+  lastOrderId: number | null;
+  lastOrderShortAlias: string | null;
+  activeOrderId: number | null;
+  activeOrderShortAlias: string | null;
   order: Order | null;
 }
```
KoalaSat, line 11: This is the most important change, everything else is just adapting the code to it

KoalaSat:
It works, but I think I need to review the order concurrent fetcher on `FederationContext`

Reckless-Satoshi:
Hey @KoalaSat , I have only been able to skim through the code and I have not run it myself. However it looks really good! Getting rid of the redundant `currentOrder` was a must. Let's merge it as is to the v0.6.0 branch so I am not a block on this great progress!

PR #955, opened 2023-11-17
Change getRobot by getSlot

KoalaSat:
## What does this PR do?
There is a big difference between:
- `garage.getRobot().active_order_id`: order id of the last coordinator in return robot information
- `garage.getSlot().ActiveOrderId`: order id of an active order in one of the coordinators

I just removed the possibility to directly access to the robot through the garage and force to always land first into the slot  (`garage.getSlot().robot.found`) so it's clear where the aggregated information is (`garage.getSlot().ActiveOrderId`).

## Checklist before merging

### DIFF
--- a/frontend/src/basic/BookPage/index.tsx
+++ b/frontend/src/basic/BookPage/index.tsx
@@ -31,7 +31,7 @@ const BookPage = (): JSX.Element => {
   const chartWidthEm = width - maxBookTableWidth;
 
   const onOrderClicked = function (id: number, shortAlias: string): void {
-    if (garage.getRobot().avatarLoaded) {
+    if (garage.getSlot().robot.avatarLoaded) {
       setDelay(10000);
       navigate(`/order/${shortAlias}/${id}`);
     } else {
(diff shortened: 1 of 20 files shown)


KoalaSat:
Maybe at some point, we should review this and store the same robot information separated by coordinator, with a actual status, a bad coordinator might introduce wrong robot data and break the front-end.

Reckless-Satoshi:
> robot information separated by coordinator,

Yes this is a must. I was envisioning `robot` to be one more attribute to the `Coordinator` model, however, that is difficult to reconcile with the Garage.

KoalaSat:
> that is difficult to reconcile with the Garage.

Maybe storing in the garage slot the specific data from each coordinator? That woudl also simplify the actual Slot entity

```
slot = {
  token: 123,
  activeCoordinator: 'exp'
  lastCoordinator: 'satstralia',
  coordinatorActivity: {
    'exp': { ... robot ... }
    'satstralia': { ... robot ... }
  }
}
```

Reckless-Satoshi:
> activeCoordinator: 'exp'
>   lastCoordinator: 'satstralia',
>   coordinatorActivity: {
>     'exp': { ... robot ... }
>     'satstralia': { ... robot ... }

Ouh yes! This is probably the most elegant solution!

Reckless-Satoshi:
Merging to v0.6.0 as is! Thank you @KoalaSat

PR #957, opened 2023-11-18
Fix order workflow

KoalaSat:
## What does this PR do?
Fixes the order page and refactor tons of code:
- Remove the `focusedCoordinator` concept ,it had to be updated and maintained and was just being used to obtain the `shortAlias` for the Coordinator dialog, I just changed the boolean to string in `OpenDialogs`. It was also being used by the MakerComponent, but the maker entity already stores the coordinator info :)
- Splits the `currentOrder` concept in `FederationContext` into:
  - `focusedOrder`: Based on the params when visiting `OrderPage`, it's fetched just once and only used to populate `OrderDetails`
  - `garage.getSlot().garage` it's the once being updated in `FederationContext` for notifications and populate `TradeBox`
- Because of that, clearly separates the data in `OrderDetails` from the one in `TradeBox`.
- Changed the autoupdate in `FederationContext`, I think before it was working just by chance
- Stop passing values to internal components that could be obtained directly from the contexts
- Refactor some duplicated code

## Checklist before merging


frontend/src/basic/MainDialogs/index.tsx
```diff
@@ -56,7 +45,7 @@ const MainDialogs = (): JSX.Element => {
       />
       <AboutDialog
         open={open.info}
-        maxAmount={maxAmount}
+        maxAmount={'100000'} //FIXME About dialog shoul allow to change coordinator
```
KoalaSat, line 48: ⚠️

Reckless-Satoshi: About dialog has to change for the federation. It can be simplified, as every coordinator has its own terms.

PR #961, opened 2023-11-21
Adapt garage to multi coordinators

KoalaSat:
## What does this PR do?
Implemented `Slot` entity to be able to store robot data from different coordinators under the same token:
- Moved `token`, `avatarLoaded`and `copiedToken` at Slot level
- Changed the slot array in garage to a map `{token => Slot}`
- Implement `getRobot()` to access specific robots from different coordinators in a `Slot`, no more direct access like `getSlot().robot`
- Changed `localsotage` key since the entity is completely different now

## Checklist before merging

### DIFF
--- a/frontend/src/basic/BookPage/index.tsx
+++ b/frontend/src/basic/BookPage/index.tsx
@@ -31,7 +31,7 @@ const BookPage = (): JSX.Element => {
   const chartWidthEm = width - maxBookTableWidth;
 
   const onOrderClicked = function (id: number, shortAlias: string): void {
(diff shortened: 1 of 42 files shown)


Reckless-Satoshi:
Thank you for these refactors! Works well to me. We should refrain for some time from more large changes to `the-federation-layer-v0.6.0` otherwise we keep will keep breaking things :D

KoalaSat:
@Reckless-Satoshi I believe this is the last big change needed, from now on I would only expect bugs fixing :)

PR #962, opened 2023-11-22
Update docs to point to new LND node

ltcarbonell:
This PR updates the documentation to point to the current LND node.

## Checklist before merging


Reckless-Satoshi (approved):

Reckless-Satoshi:
Hey @ltcarbonell good catch! 

It's interesting that we have a global variable `node_id` in `/docs/_config.yml` that not only, we did not use here, but in addition, is also not up to date. It goes for demonstrating how our Docs are not being payed enough attention :`)

PR #971, opened 2023-11-26
Add robo-identity generator

Reckless-Satoshi:
## What does this PR do?
Adds client side robot avatar and nickname generator. Only on Web, breaks the Android react-native client given that it requires webassembly and webworkers. We have to compile the Rust libraries for Android target and run them from Java.

WIP

## Checklist before merging

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -47,6 +47,7 @@
         "react-smooth-image": "^1.1.0",
         "react-world-flags": "^1.6.0",
         "reconnecting-websocket": "^4.4.0",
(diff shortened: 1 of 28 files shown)


Reckless-Satoshi:
Working but not fully working. Only works "sometimes", it might have to do with how `hashId` changes are not detected and rerendered. What do you think @KoalaSat ? I'll mark it as ready for review.

Reckless-Satoshi:
Working but not fully working. Only works "sometimes", it might have to do with how `hashId` changes are not detected and rerendered. What do you think @KoalaSat ? I'll mark it as ready for review.

PR #973, opened 2023-11-27
fix: kycnot.me url

alexandresanlim:
This PR fix kycnot.me url review

## Checklist before merging


Reckless-Satoshi:
Good catch, thank you for fixing this link

PR #1003, opened 2023-12-15
wallets compatibility - add breez

alanStocco:
## What does this PR do?
Add in wallets compatibility page Breez mobile wallet.
I tried it some times as buyer without problem. Never tried as seller but I read some good review in telegram.


alanStocco:
Sorry wrong Total icon. I'll fix in another PR

PR #1035, opened 2023-12-30
Robohash gereator queue

KoalaSat:
## What does this PR do?
Instead of running Workers every time a robohash has to be calculated, it adds the task to a queue to avoid mixing responses issues we found in Mozilla's implementation.

## Checklist before merging

### DIFF
--- a/frontend/src/models/Federation.model.ts
+++ b/frontend/src/models/Federation.model.ts
@@ -17,7 +17,7 @@ export class Federation {
   constructor() {
     this.coordinators = Object.entries(defaultFederation).reduce(
       (acc: Record<string, Coordinator>, [key, value]: [string, any]) => {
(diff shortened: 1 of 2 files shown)


frontend/src/models/Federation.model.ts
```diff
@@ -17,7 +17,7 @@ export class Federation {
   constructor() {
     this.coordinators = Object.entries(defaultFederation).reduce(
       (acc: Record<string, Coordinator>, [key, value]: [string, any]) => {
-        if (getHost() !== '127.0.0.1:8000' && key == 'local') {
+        if (getHost() !== '127.0.0.1:8000' && key === 'local') {
```
KoalaSat, line 20: just a lintern

PR #1039, opened 2023-12-31
Fix depth chart and coordinator table

KoalaSat:
## What does this PR do?
- When switched to mainnet/testnet the coordinators table gets refreshed
- Depth chart now displays avatars

## Checklist before merging


Reckless-Satoshi (commented):
Great fixes @KoalaSat !!

PR #1064, opened 2024-01-12
Fix contexts

KoalaSat:
## What does this PR do?
Fixes #<ISSUE_NUMBER>

## Checklist before merging

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,8 +1,7 @@
 import React, { StrictMode, Suspense } from 'react';
 import ReactDOM from 'react-dom/client';
 import Main from './basic/Main';
-import { CssBaseline, ThemeProvider } from '@mui/material';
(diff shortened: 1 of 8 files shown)


Reckless-Satoshi:
Thank you very much for these critical fixes.

The `currentOrder` gives me a good intuition on where to keep pull strings. However, it is very broken (it moved the hook `useParams`, outside of a body of a React functional component and also outside of ReactRouter). I will think take the inspiration of it and find a better way to re-implement currentOrder.

The Contexts fix works great! I merged already only that commit into `the-federation-layer-v0.6.0`. The fix disclosed an infinite re-render loop bug that was in FederationContext (but never took effect before) given that `robotUpdatedAt` was a dependency of the `useEffect` that fetches robots. After fixing it, the context problem is solved. Thank you!

I believe we can close this PR.

KoalaSat:
Perfect @Reckless-Satoshi thanks for the last check, closing

PR #1086, opened 2024-01-23
Fix book loading

KoalaSat:
## What does this PR do?
More granulated exchange data to fix the book loading missfit
## Checklist before merging

### DIFF
--- a/frontend/src/models/Exchange.model.ts
+++ b/frontend/src/models/Exchange.model.ts
@@ -61,13 +61,17 @@ export const updateExchangeInfo = (federation: Federation): ExchangeInfo => {
 
 export interface Exchange {
   info: ExchangeInfo;
+  activeCoordinators: number;
   onlineCoordinators: number;
+  loadingCoordinators: number;
   totalCoordinators: number;
 }
 
 export const defaultExchange: Exchange = {
   info: defaultExchangeInfo,
+  activeCoordinators: 0,
   onlineCoordinators: 0,
+  loadingCoordinators: 0,
   totalCoordinators: 0,
 };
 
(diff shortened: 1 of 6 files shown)


frontend/src/models/Exchange.model.ts
```diff
@@ -59,13 +59,17 @@ export const updateExchangeInfo = (federation: Federation): ExchangeInfo => {
 
 export interface Exchange {
   info: ExchangeInfo;
+  activeCoordinators: number;
   onlineCoordinators: number;
+  loadingCoordinators: number;
   totalCoordinators: number;
```
KoalaSat, line 65: activeCoordinators: # of coordinators marked as active by the user
  onlineCoordinators: # of online coordinators
  loadingCoordinators: # of coordinators with a pending request
  totalCoordinators: # of registered coordinators

Reckless-Satoshi: `activeCoordinators` maybe it is best to rename this one to `enabledCoordinators` for consistency and to not confuse these with the `online` coordinators. I believe we use `active` and `online` sometime as synonyms in the UI.

frontend/src/models/Federation.model.ts
```diff
@@ -125,14 +134,22 @@ export class Federation {
...
   };
 
   enableCoordinator = (shortAlias: string): void => {
     this.coordinators[shortAlias].enable(() => {
+      this.updateActiveCoordinators();
       this.triggerHook('onCoordinatorUpdate');
     });
   };
+
+  updateActiveCoordinators = (): void => {
```
Reckless-Satoshi, line 148: For the same reason `updateActiveCoordinators` may be better named as`updateEnabledCoordinators()` and have `exchange.enabledCoordinators` instead of `exchange.activeCoordinators`.

Reckless-Satoshi (commented):
It looks great! I would rename that one variable, what do you think?

frontend/src/models/Federation.model.ts
KoalaSat: All changed! thanks!

PR #1099, opened 2024-01-29
Fix telegram link

KoalaSat:
## What does this PR do?
Links to telegram were not working because of the `@`, by removing it works again

## Checklist before merging


Reckless-Satoshi:
Good catch! Works as expected now.

Issue #1069, opened 2024-01-14
Bugs on v0.6.0 full list


Reckless-Satoshi:
**Describe the bugs**
This is a list of bug fixes needed on the federated app v0.6.0 (https://github.com/RoboSats/robosats/tree/the-federation-layer-v0.6.0) for functionality parity with v0.5.3 (i.e., a list of things that do not work as expected in v0.6.0 but used to work on v0.5.3).

- [ ] **maybe fixed?** There seems to be an issue very rare where sometimes, after locking the bond, the `ErrorBoundary` is triggered and a `Borked app` page is show. Seems to be related to an `Order` object where `.price` is not found (null not expected).
- [ ] **needs research** The avatar generator webworker/wasm might stop working on a browser tab refresh.
- [ ] **needs tests, is it still reproducible?** There seems to be some mishap with PGP keys across different robots on the same garage. To find this issue showing up: 1) Create a robot and create an order. 2) Open a new browser tab. 3)Create another robot, take the previous order. 4) Follow the trade until chat. 5) message each other in the chat. You will see one of the two robots (or both) cannot validate the peer's signatures on the chat messages. In fact, the peer is signing the messages with the wrong private key it seems. Weird!
- [ ] **needs tests, is it still reproducible?** It seems a second robot in the Garage cannot create an order. "You are already a maker of an active order". That is a backend response, meaning, the header of that request was sent with the wrong robot identity (the other robot's identity). This and the previous bulletpoint probably have a common cause. Maybe the `slot` keeps the info of the previous robot (pgp keys, base91 token, etc).
- [x] Aggregated summary stats behaves weird. The aggregation is broken. Sometimes is stuck at 0 values. It does not re-trigger on mainnet/testnet switch click. Does not trigger a `updateCoordinators()` when the ExchangeSummary dialog is opened (as it used to be on v0.5.4). Last 24h average premium is sometimes `NaN%`.
- [x] Sometimes the `Order` tab is not clickable even when the Robot has an active order.
- [x] Sometimes, renewing an order (i.e., creating a new order with same params as one that expired) will not show you the order just created (it will not trigger a reload, probably because of a very large `delay` between requests). However leaving the Order tab and coming back will work.
- [x] ProfileDialog loading animation does not work. Similar issue to order book loading bar. There should be an indicator of waiting for `/api/robot` requests, same as when generating a robot, everytime the profile is opened.
- [x] Deleting a custom payment method in the `MakerForm` crashes the client.
[...]

PR #1173, opened 2024-03-10
Fix undefined prices error

KoalaSat:
## What does this PR do?
Related to #1069 Item 1??

From the issue:
>  There seems to be an issue very rare where sometimes, after locking the bond, the ErrorBoundary is triggered and a Borked app page is show. Seems to be related to an Order object where .price is not found (null not expected).

An user reported a similar error when opening the `/offers` view, after checking when this can be possible (`Cannot read properties of undefined (reading 'price')` this has to happen on a function where `.price` is the first call to an `undefined` instance) and triangulating it also witb the locking bond action, I believe this fixes the same case. I'm specially motivated by the `baseL == null`, which seems to be a missed fix of the PR about the lintern errors.

## Checklist before merging

### DIFF
--- a/frontend/src/models/Limit.model.ts
+++ b/frontend/src/models/Limit.model.ts
@@ -9,7 +9,7 @@ export interface Limit {
 export type LimitList = Record<string, Limit>;
 
 export const compareUpdateLimit = (baseL: Limit, newL: Limit): Limit => {
-  if (baseL == null) {
+  if (!baseL) {
     return newL;
   } else {
     const price = (baseL.price + newL.price) / 2;


KoalaSat:
> Maybe it simply takes too long to reply and it feels like it crashed randomly?

Maybe? It's possible yeah


```
lnbc200u1pjl9uuzpp5qtx5u332jc78ty3xm46a9jsgmdj75qn7kn2saclt2z3ysyp6c5usdqqcqzzsxqy8ayqsp54stsy8qr5hc4dx4usxp7wzuvu9x36vl9us6hx25rywcnccla90ms9qyyssqdvqkdaucfvhq3s84qxt6hp2qsj2jdhg4k26cq28dtedtl5mc6mxp00zm2wvufyg9ka072fgyq03n50fsuyv3nr4f2yxt2lw6msmrspsp39x602
```

Reckless-Satoshi:
> lnbc200u1pjl9uuzpp5qtx5u332jc78ty3xm46a9jsgmdj75qn7kn2saclt2z3ysyp6c5usdqqcqzzsxqy8ayqsp54stsy8qr5hc4dx4usxp7wzuvu9x36vl9us6hx25rywcnccla90ms9qyyssqdvqkdaucfvhq3s84qxt6hp2qsj2jdhg4k26cq28dtedtl5mc6mxp00zm2wvufyg9ka072fgyq03n50fsuyv3nr4f2yxt2lw6msmrspsp39x602

823a7d69d2081985249efdc12defbc0094d496ee184b7b79896e4492a18f84ca

PR #1195, opened 2024-03-20
Italian translation

lechatnieger:
## What does this PR do?
Fixes the italian translation.


frontend/static/federation.json
```diff
@@ -1,6 +1,6 @@
 {
   "exp": {
-    "longAlias": "Experimental",
+    "longAlias": "BitcoinVeneto",
```
Reckless-Satoshi, line 3: Can you remove this change?

PR #1260, opened 2024-05-01
Fix hard coded receiving method into order detail message.

rodrigodh:
# What does this PR do?
This PR introduces a fix for the order detail box about the receiving payment message, so users can have it more clear when they are receiving through on-chain or lightning.

I found this problem on my last transaction when even though i selected on-chain it was showing lightning, but was just a visual issue.

Screenshot 2024-05-01 at 12 28 51

## Checklist before merging

It's important to notice that I don't have much experience with open-source or crypto, so I'm up to fix any mistake on my part.

### DIFF
--- a/frontend/src/components/OrderDetails/index.tsx
+++ b/frontend/src/components/OrderDetails/index.tsx
@@ -207,7 +207,7 @@ const OrderDetails = ({
         amount: amountString,
         method: order.payment_method,
       });
-      receive = t('You receive via Lightning {{amount}} Sats (Approx)', {
+      receive = t('You receive {{amount}} Sats (Approx)', {
         amount: sats,
       });
     } else {
--- a/frontend/static/locales/ca.json
+++ b/frontend/static/locales/ca.json
@@ -464,7 +464,7 @@
   "The order has expired": "L'ordre ha expirat",
   "The pinned location is approximate. The exact location for the meeting place must be exchanged in the encrypted chat.": "The pinned location is approximate. The exact location for the meeting place must be exchanged in the encrypted chat.",
   "You cannot take an order yet! Wait {{timeMin}}m {{timeSec}}s": "Encara no pots prendre cap ordre! Espera {{timeMin}}m {{timeSec}}s",
-  "You receive via Lightning {{amount}} Sats (Approx)": "Tu reps via Lightning {{amount}} Sats (Approx)",
+  "You receive {{amount}} Sats (Approx)": "Tu reps via Lightning {{amount}} Sats (Approx)",
   "You receive via {{method}} {{amount}}": "Reps via {{method}} {{amount}}",
   "You send via Lightning {{amount}} Sats (Approx)": "Tu envies via Lightning {{amount}} Sats (Approx)",
   "You send via {{method}} {{amount}}": "Envies via {{method}} {{amount}}",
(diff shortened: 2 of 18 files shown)


frontend/src/components/OrderDetails/index.tsx
```diff
@@ -207,8 +207,10 @@ const OrderDetails = ({
         amount: amountString,
         method: order.payment_method,
       });
-      receive = t('You receive via Lightning {{amount}} Sats (Approx)', {
+      const receives_on_chain = order.swap_fee_rate > 0;
```
Reckless-Satoshi, line 210: Is `order.swap_fee_rate` the best param we can use for this? This might display "onchain" for everyone while they are in the step of submitting an invoice.

Ideally we have a `order.is_swap` boolean that is only sent to the buyer.

Reckless-Satoshi (commented):
Good catch. We might need to expose a better param from the API for this to work. The alternative is to delete the `via {method}` so it will simply read "You receive XX Sats (Approx)"

frontend/src/components/OrderDetails/index.tsx
rodrigodh: Since we don't have this param from the API now, i just changed the text.

Reckless-Satoshi (approved):
Thank you @rodrigodh ! It is not the best possible solution, but at least it is not confusing anymore (that's what matters).

Congratulations on your first contribution. Please submit an invoice for 15K Sats for a small reward from the Development fund.

rodrigodh:
> Thank you @rodrigodh ! It is not the best possible solution, but at least it is not confusing anymore (that's what matters).
> 
> Congratulations on your first contribution. Please submit an invoice for 15K Sats for a small reward from the Development fund.

Thank you! I'm not sure if my Lightning Network is set up correctly, but you can try this invoice. (don't worry if it doesn't work.)

I hope I can contribute even more in the future.

```
lnbc150u1pnr9674pp55jmsugjuzxmhv777rk44stqt9l27mc64w4a8kqjc6a7dmvvlk4lqsp5u5g0vt4jadcqnsmnsg7hzdex32dpap8dhxwxl3cauem9rqa52q5qdq2ga5hg6r4vgcqzynxqyjw5q9qy9scqfppqmqttp9zvyz5avpc42e4xh53m4z4hlj79rzjqvv2e8a2jc570ksgsxdu3lsd62hrq3xknvdj9qaxx3u6elltj6zgxrx03vqqp9gqqgqqqqlgqqqqqqgq2q9lmwdq4kufmu0pun6hleqjpgc7655yptfwyarrgerqlw662stymskd3wgv86nnl5ddnl0s76khhnts2hm6794sucxcuzhwrqr795y8cp8qeaxr
```

rodrigodh:
@Reckless-Satoshi is it merging?

Reckless-Satoshi:
> lnbc150u1pnr9674pp55jmsugjuzxmhv777rk44stqt9l27mc64w4a8kqjc6a7dmvvlk4lqsp5u5g0vt4jadcqnsmnsg7hzdex32dpap8dhxwxl3cauem9rqa52q5qdq2ga5hg6r4vgcqzynxqyjw5q9qy9scqfppqmqttp9zvyz5avpc42e4xh53m4z4hlj79rzjqvv2e8a2jc570ksgsxdu3lsd62hrq3xknvdj9qaxx3u6elltj6zgxrx03vqqp9gqqgqqqqlgqqqqqqgq2q9lmwdq4kufmu0pun6hleqjpgc7655yptfwyarrgerqlw662stymskd3wgv86nnl5ddnl0s76khhnts2hm6794sucxcuzhwrqr795y8cp8qeaxr

I get "No Route". Please feel free to post another one with a long expiration time so I can keep retrying.

Merging this PR :rocket:

PR #1297, opened 2024-05-17
Add external coordinator

KoalaSat:
## What does this PR do?
Adds the option of including an external coordinator

image

Can be easily tested in local by removing an existing coordinator form `frontend/static/federation.json` and try to add it as external.

**WARNING**: This PR touches all important models so it requires good testing

## Checklist before merging


frontend/src/components/FederationTable/index.tsx
```diff
@@ -76,7 +77,8 @@ const FederationTable = ({
           >
             <Grid item>
               <RobotAvatar
-                shortAlias={params.row.shortAlias}
+                shortAlias={federated ? params.row.shortAlias : undefined}
+                hashId={federated ? undefined : params.row.shortAlias}
```
Reckless-Satoshi, line 81: NIT: for external coordinators we could use their `mainnet.url` instead of the `shortAlias` for the hashId. This way we ensure that the avatar generated matches the URL (permanent) instead of the alias (user dependentant) therefore the avatar will always be the same for everyone who adds this coordinator.

Reckless-Satoshi (commented):
This is great Koala!

I am having a couple issues. 

On localhost (127.0.0.1:8000) this PR makes the list of coordinators look like this (some are missing, including "Local dev" and should show only under this host):
image

Adding coordinators this way will only work on Tor Browser and Android. However, this will not work on the selfHosted app (because for it to work, a new socat bridge is needed on the node to expose that coordinator's Onion locally). Maybe we can write a condition to hide those UI elements in the selfHosted app?

I think I need to test this further before we merge :D

KoalaSat:
@Reckless-Satoshi All changes done

frontend/src/components/BookTable/BookControl.tsx
```diff
@@ -359,7 +359,8 @@ const BookControl = ({
                     >
                       <div style={{ display: 'flex', alignItems: 'center', flexWrap: 'wrap' }}>
                         <RobotAvatar
-                          shortAlias={coordinator.shortAlias}
+                          shortAlias={coordinator.federated ? coordinator.shortAlias : undefined}
+                          hashId={coordinator.federated ? undefined : coordinator.shortAlias}
```
Reckless-Satoshi, line 363: The `coordinator.mainnet.onion` fix is needed here too so the avatar is the same as in the `BookTable`

frontend/src/components/OrderDetails/index.tsx
```diff
@@ -265,7 +265,12 @@ const OrderDetails = ({
             {' '}
             <Grid container direction='row' justifyContent='center' alignItems='center'>
               <Grid item xs={2}>
-                <RobotAvatar shortAlias={coordinator.shortAlias} small={true} smooth={true} />
+                <RobotAvatar
+                  shortAlias={coordinator.federated ? coordinator.shortAlias : undefined}
+                  hashId={coordinator.federated ? undefined : coordinator.shortAlias}
```
Reckless-Satoshi, line 270: Same here
```
hashId={coordinator.federated ? undefined : coordinator.mainnet.onion}
```

frontend/src/components/MakerForm/SelectCoordinator.tsx
```diff
@@ -83,7 +79,8 @@ const SelectCoordinator: React.FC<SelectCoordinatorProps> = ({
             >
               <Grid item>
                 <RobotAvatar
-                  shortAlias={coordinatorAlias}
+                  shortAlias={coordinator?.federated ? coordinator.shortAlias : undefined}
+                  hashId={coordinator?.federated ? undefined : coordinator.shortAlias}
```
Reckless-Satoshi, line 83: And here

KoalaSat: everything changed, thank you for noticing!

Reckless-Satoshi (approved):
I did not test once again, but I'm gonna trust it just works as intended looking at the code! 😄 🚀

PR #1318, opened 2024-06-07
Notifications Endpoint

KoalaSat:
## What does this PR do?
This PR introduces a new API endpoint for the robot to fetch all notifications sent to it. To reduce bandwidth it allows to filter by the creation date. 

A new table `Notification` has been created to have a place to store these notifications.

## Checklist before merging


api/management/commands/telegram_watcher.py
```diff
@@ -6,7 +6,7 @@
 from django.db import transaction
 
 from api.models import Robot
-from api.notifications import Telegram
+from api.notifications import Notifications
```
KoalaSat, line 9: I think this change makes sense

api/notifications.py
```diff
@@ -3,10 +3,11 @@
 from decouple import config
 
 from api.models import Order
+from api.models import Notification
 from api.utils import get_session
 
 
-class Telegram:
+class Notifications:
```
KoalaSat, line 10: I think this change makes sense now that we have several ways to notificate

api/notifications.py
```diff
@@ -29,13 +30,25 @@ def get_context(user):
...
+        notification.description = description
+        notification.user = robot
+        notification.save()
+
+    def send_telegram_message(self, chat_id, title, description):
         """sends a message to a user with telegram notifications enabled"""
 
         bot_token = config("TELEGRAM_TOKEN")
-
+        text = f"{title} {description}"
```
KoalaSat, line 50: I want to use just the title part for Android notifications

KoalaSat:
Moved to https://github.com/RoboSats/robosats/pull/1347

PR #1332, opened 2024-06-16
Fix sats text on range orders

KoalaSat:
## What does this PR do?

When a order with a range was taken, it shows always the value in sats for the max range, regardless the amount chosen

## Checklist before merging


frontend/src/components/OrderDetails/index.tsx
```diff
@@ -178,7 +178,7 @@ const OrderDetails = ({
       : coordinator.info?.taker_fee ?? 0;
     const defaultRoutingBudget = 0.001;
     const btc_now = order.satoshis_now / 100000000;
-    const rate = order.amount > 0 ? order.amount / btc_now : Number(order.max_amount) / btc_now;
+    const rate = Number(order.max_amount ?? order.amount) / btc_now;
```
KoalaSat, line 181: `btc_now` is calculated over `max_amount` if exists

Reckless-Satoshi: I am assuming this has been tested. Based on the change of logic alone I am not sure what the result would be (when is `max_amount` `null | undefined`? no idea), but looks like this probably fixes the issue :smile: Merging as is!

PR #1333, opened 2024-06-16
Update order navbar on time

KoalaSat:
## What does this PR do?
When switching robots, sometimes the navbar was not being synchronized in time, leading to bad urls

## Checklist before merging


KoalaSat:
Tested in local, merging as it is 👍

PR #1347, opened 2024-06-21
Notifications api endpoint

KoalaSat:
## What does this PR do?
This PR introduces a new API endpoint for the robot to fetch all notifications sent to it. To reduce bandwidth it allows to filter by the creation date.

A new table Notification has been created to have a place to store these notifications.

## Checklist before merging


docs/assets/schemas/api-latest.yaml
```diff
@@ -284,6 +284,30 @@ paths:
                     type: string
                     description: Reason for the failure
           description: ''
+  /api/notifications/:
```
Reckless-Satoshi, line 287: By running the tests the API schema and redoc docs get updated with thr new endpoint 👌

tests/utils/trade.py
```diff
@@ -111,6 +111,15 @@ def get_order(self, robot_index=1, first_encounter=False):
...
         self.response = self.client.get(path + params, **headers)
 
+    def get_notifications(self, created_at=0, robot_index=1, first_encounter=False):
+        """
+        Fetch the latest state of the order
+        """
+        path = reverse("notifications")
+        params = f"?created_at={created_at}"
+        headers = self.get_robot_auth(robot_index, first_encounter)
+        self.response = self.client.get(path + params, **headers)
```
Reckless-Satoshi, line 121: During tests, `trade.response` is always the return of the latest GET to `/api/order`. Maybe here we want to create a new variable on the Trade class `notification` . This way, in tests we evaluate the return of `trade.response` and of `trade.notification` and don't get confused when improving them in the future. 

I would therefore edit this one as:
```
self.notification = self.client.get(path + params, **headers)
```
`notification` or `last_notification` , the one name you prefer!

tests/test_trade_pipeline.py
```diff
@@ -802,6 +955,17 @@ def test_taken_order_expires(self):
 
         self.assert_order_logs(data["id"])
 
+        trade.get_notifications()
+        # Checks
+        self.assertResponse(trade.response)
```
Reckless-Satoshi, line 960: If we use `trade.notification` as the response stored by the get_notification() method, that will make tests more clear :)

Reckless-Satoshi (commented):
Thank you for adding these tests! This looks very ready. Left just a couple NIT comments :D

KoalaSat:
Sorry for the confusion @Reckless-Satoshi 😆 test are not yet ready to be review, I'm investigating a possible wrong test on `main`

tests/utils/trade.py
KoalaSat: I ended up moving it out of Trade, I think it didn't make sense to have it there

KoalaSat:
https://github.com/RoboSats/robosats/actions/runs/9686642409

PR #1353, opened 2024-06-25
Fix RegExp

KoalaSat:
## What does this PR do?
Fixes regexp to not allow urls with no porotcol

## Checklist before merging


KoalaSat (commented):
Tested and merged

PR #1354, opened 2024-06-26
fix verify_signed_message and tests/robots/1/signed_message

jerryfletcher21:
## What does this PR do?
`verify_signed_message` was not checking for the validity of the signature. The documentation of `python-gnupg` is not clear about this, it says that the fields returned by `verify` are set just if the signature is valid. In this case `tests/robots/1/signed_message` was signed with the correct key but with the wrong digest-algo (`SHA256` instead of `SHA512`) as expressed on the signature.
Running `gpg --verify tests/robots/1/signed_message` after importing the key returned: 
```
gpg: WARNING: signature digest conflict in message
gpg: Can't check signature: General error
```
and a non zero error code. Changing `Hash: SHA256` in `tests/robots/1/signed_message` returned no errors.

`verify_signed_message` called in `test_verify_signed_message` in `api/tests/test_utils.py` was not catching this because `verified.fingerprint` was set even though `verified.valid` was `False`.

I have recreated the signature with `SHA512` by running:
```
gpg --pinentry-mode loopback --passphrase-file tests/robots/1/token --import tests/robots/1/enc_priv_key

cat tests/robots/1/signed_message | head -n 4 | tail -n 1 | gpg --armor --digest-algo SHA512 --clearsign --local-user "$(gpg --show-key tests/robots/1/pub_key | head -n 2 | tail -n 1)" --pinentry-mode loopback --passphrase-file tests/robots/1/token > tests/robots/1/signed_message
```

## Checklist before merging


Reckless-Satoshi:
Thanks for figuring this super sneaky one @jerryfletcher21 !

This PR needs a lot of testing as we risk getting all invoices submitted rejected.

KoalaSat:
Test passing https://github.com/RoboSats/robosats/actions/runs/11418567681

KoalaSat:
I managed to run an entire trade workflow with https://hub.docker.com/layers/recksato/robosats/9e442c3/images/sha256-64343a7dc5b7c7b8805fc9e60ee8d74020c34da4a028a5c02ca492d48d9ce81a?context=explore

PR #1360, opened 2024-06-30
Fix build-it in

KoalaSat:
## What does this PR do?
Fixes build=it selection code

## Checklist before merging


KoalaSat:
Tested in 2 devices with no orbot and orbot

PR #1363, opened 2024-07-04
Split host and port in .env variables

jerryfletcher21:
## What does this PR do?
Split host and port in .env variables

## Checklist before merging


KoalaSat (commented):
Nice proposal! We shouldn't forget about double check if there is any effect on https://github.com/RoboSats/robosats-deploy, since its the software that coordinators run

jerryfletcher21:
@KoalaSat this PR is related to #1364, but I split them in two different PR to better manage them. Ah yes thank you I did not think about robosats-deploy, I will have to check it.

jerryfletcher21:
@KoalaSat I have made a [PR](https://github.com/RoboSats/robosats-deploy/pull/15) also in robosats-deploy, thank you for your feedback!

Reckless-Satoshi:
If we want to add new variables for host/port split into two vars, we need to keep supporting the current variable names with deprecation notice. Re-factoring the `.env` can lead to production issues to the coordinators as they need to conciously go and edit their `.env` (in their robosats-deploy) at the moment they pull the new robosats coordinator image.

What is the motivation for splitting host and port into two variables? I agree it's cleaner, but not sure if worth the amount of trouble this upgrade can potentially cause.

jerryfletcher21:
@Reckless-Satoshi yes the coordinators would have to change their `.env` files and I agree it may lead to production issues if they are not informed correctly. 
I made this PR because while developing #1364, I needed to get just the port and not the host, and also because postgres variables in `.env` are already separated in host and port, so it is cleaner and more modular to have them all separated.
I could rewrite this PR to still support the old variables if the new ones are not present, or if you do not want the trouble this PR may cause I can simply extract the ports from the variables in #1364 and not base it on this PR.

robosats/celery/__init__.py
```diff
@@ -3,11 +3,15 @@
...
+from decouple import config
 from celery import Celery
 from celery.schedules import crontab
 
 # You can use rabbitmq instead here.
-BASE_REDIS_URL = os.environ.get("REDIS_URL", "redis://localhost:6379")
+BASE_REDIS_URL = \
+    "redis://" + str(config("REDIS_HOST", cast=str, default="localhost")) + \
+    ":" + str(config("REDIS_PORT", cast=str, default="6379")) + \
+    "/" + str(config("REDIS_DB_NUMBER", cast=str, default="1"))
```
KoalaSat, line 14: Seems like REDIS_DB_NUMBER was not present previously

PR #1364, opened 2024-07-04
Traditional non docker environment for backend development

jerryfletcher21:
## What does this PR do?
Builds on #1363.
This PR adds a setup for a traditional non docker regtest environment for backed development and testing.
[scripts/traditional/README.md](https://github.com/RoboSats/robosats/blob/15327289a9f512dca8772d5efed9ac9dd2f50dd2/scripts/traditional/README.md) with all the informations.

## Checklist before merging


Reckless-Satoshi:
This is super useful! What about we base the `tradition-environment` over `main` instead of `split-host-port` given how traumatic that fix is?

jerryfletcher21:
> This is super useful! What about we base the `tradition-environment` over `main` instead of over `split-host-port` given how traumatic that fix is?

It would be a bit less clean since we would have to extract the port from the variables, but it is of course straightforward to do.

jerryfletcher21:
Closing this in favor of #1940

PR #1372, opened 2024-07-08
Add TOTP to admin panel

KoalaSat:
## What does this PR do?
Includes TOTP as part of the admin login, admin users with no TOTP associated can still login as normal

## Checklist before merging


robosats/urls.py
```diff
@@ -15,19 +15,30 @@
...
+from django_otp.admin import OTPAdminSite
+
+
+class OTPAdmin(OTPAdminSite):
+    pass
+
+
+admin_site = OTPAdmin(name="OTPAdmin")
+for model in apps.get_models():
+    admin_site.register(model)
```
KoalaSat, line 31: This includes more (hidden) models than the normal admin view, but honestly I found them useful

PR #1374, opened 2024-07-11
script to generate latest api specs file

jerryfletcher21:
## What does this PR do?
Right now [docs/assets/schemas/api-latest.yaml](https://github.com/RoboSats/robosats/blob/main/docs/assets/schemas/api-latest.yaml) is updated just when running the tests in [tests/test_api.py](https://github.com/RoboSats/robosats/blob/main/tests/test_api.py) if `python3 manage.py runserver` is running. The tests can be run without the django server `python3 manage.py runserver` running, but `api-latest.yaml` will not be updated.
This script adds a way to update `api-latest.yaml` from a running django server without having to run the tests.

## Checklist before merging

### DIFF
--- a/scripts/api-specs-update.py
+++ b/scripts/api-specs-update.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python3
+
+"""Update the latest api specs file from a running django server"""
+
+import sys
+
+import urllib.request
+import urllib.error
+
+
+try:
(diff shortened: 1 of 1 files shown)


Reckless-Satoshi:
This is veery useful! Thanks a lot @jerryfletcher21

PR #1380, opened 2024-07-12
Android notification system

KoalaSat:
## What does this PR do?

Screenshot from 2024-07-12 21-48-50

This PR includes all the necessary code to have torified notifications. A background job runs every 5 minutes, even if the app is off, to check if the user has new notifications on their active orders. Important points to consider:

- There is a warning notification that runs every time the app starts, I couldn't find a way to show it only once, since it appears to be a warning forced by android.
- The 5 minutes gap is set to avoid DDoSing our own coordinators, the idea is still to allow people to run coordinators with minimum reliable machines from home.
- For the same reason, the android app has to see at least once the order of the robot, that way we make sure it's worth to query. Doing this for every robot in the garage for all coordinators every 5 minuter will be also dead by DDoS.
- If the user taps on the notification, the app opens with the order view even if it was off. I had to do some black magic here because of WebView so it's worth testing.

This code also takes into consideration https://github.com/RoboSats/robosats/pull/1356, being compatible with latest release's code and these changes.

## Checklist before merging


frontend/src/basic/Routes.tsx
```diff
@@ -0,0 +1,114 @@
...
+      const coordinator = event?.detail?.coordinator;
+      if (orderId && coordinator) {
+        const slot = garage.getSlotByOrder(coordinator, orderId);
+        if (slot?.token) {
+          garage.setCurrentSlot(slot?.token);
+          navigate(`/order/${coordinator}/${orderId}`);
+        }
+      }
+    });
+  }, []);
```
KoalaSat, line 27: Necessary to catch navigation requests form the system

frontend/src/models/Federation.model.ts
```diff
@@ -90,9 +91,12 @@ export class Federation {
   };
 
   updateUrl = async (origin: Origin, settings: Settings, hostUrl: string): Promise<void> => {
+    const federationUrls = {};
     for (const coor of Object.values(this.coordinators)) {
       coor.updateUrl(origin, settings, hostUrl);
+      federationUrls[coor.shortAlias] = coor.url;
     }
+    systemClient.setCookie('federation', JSON.stringify(federationUrls));
```
KoalaSat, line 99: Informs the system the current coordinators and their URLs, useful when changing to tesnet and to avoid duplication on the system

mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,9 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
 
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
```
KoalaSat, line 7: New permissions to run notifications

mobile/android/app/src/main/java/com/robosats/tor/TorKmpManager.kt
```diff
@@ -387,3 +388,16 @@ class TorKmpManager(application : Application) {
...
+
+    @Throws(UninitializedPropertyAccessException::class)
+    fun getTorKmpObject(): TorKmp {
+        return torKmp
+    }
+
+    fun updateTorKmpObject(newKmpObject: TorKmp) {
+        torKmp = newKmpObject
+    }
+}
```
KoalaSat, line 403: You can´t have 2 TorKmp running on your app, so this singleton makes sure it's accessible from the app and the notification foreground service

mobile/patch_modules/react-native-tor/android/build.gradle
```diff
@@ -1,132 +0,0 @@
-buildscript {
```
KoalaSat, line 1: cleaning the house

api/lightning/pip-build-tracker-bjlgta7z/8bdde18590295c5842ccad049a1ac390714fb6975e494a7bbd1455e6
Reckless-Satoshi, line 1: Looks like we might have commited a few files under `api/lightning` that are not part of this PR and possibly shouod be ignored?

KoalaSat: I have no idea why but these files has been extremely persistent on being created, everytine I tried to do a commit they appear probably because after-commit, any clue?

KoalaSat: Removed anyways

KoalaSat:
Tested and merging after some users giving good feedback

PR #1389, opened 2024-07-17
Fix telegram bot

KoalaSat:
## What does this PR do?
If the TG bot finds this error, it just stays broken because every time it tries to run for the queued orders it breaks

## Checklist before merging


KoalaSat:
Merging this fix as it was confirmed to work in coordinators

PR #1460, opened 2024-09-12
Fix self-hosted client image

KoalaSat:
## What does this PR do?
Fixes the self-hosted image generation

## Checklist before merging


KoalaSat:
Tested for umbrel https://github.com/getumbrel/umbrel-apps/pull/1462

PR #1613, opened 2024-11-11
sort cheaper offers first

fiatjaf:
I've tried to run it locally but failed, so I don't know if this works.

### DIFF
--- a/frontend/src/components/BookTable/index.tsx
+++ b/frontend/src/components/BookTable/index.tsx
@@ -919,6 +919,11 @@ const BookTable = ({
         }
       >
         <ClickThroughDataGrid
+          initialState={{
+            sorting: {
+              sortModel: [{ field: 'premium', sort: 'asc' }],
+            },
+          }
           sx={headerStyleFix}
           localeText={localeText}
           rowHeight={3.714 * theme.typography.fontSize}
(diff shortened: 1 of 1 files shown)


KoalaSat:
@fiatjaf It's cheap in case you are buyer, but not if you are seller :smiley: 

Maybe we can automatically modify that sorting if you filter by SELL or BUY?

fiatjaf:
Anything is better than sorting by random or by whatever it is today.

Reckless-Satoshi:
Indeed, we should improve the sorting.

I believe ordering by coordinator probably has no effect on the willingness to coordinate, but sorting by coordinator is the only "perk" coordinators get when they donate some sats to the development of the project (TLDR; slightly higher chance to show on top correlates to their stated voluntary devfund contribution. Coordinators that donate 0 will still show on top, it is just less likely.).

In order to not break coordinator expectations, what about we first try hierarchical sorting. Factor 1. order host, factor 2. cheapness (accounting for whether it's a "buy" or "sell"). Maybe overly complicated...?

fiatjaf:
Honestly the first time I browsed this I was so confused because I assumed it was sorted by price and then I only saw some very high prices and I thought the liquidity of the app was awful because the best price options were all on the third page, so I opened this.

I didn't know this was by design as a perk for coordinators. But thinking now I don't think this makes sense as a perk either.

A better perk would be to default the coordinator at order creation time to the top coordinators. That will probably have a much bigger impact on coordinator usage.

Maybe you're doing this already? How do you select what is the default coordinator when creating an order?

If you have coordinators order by total contributions, like

1. Coordinator A: 60% of contributions
2. Coordinator B: 30% of contributions
3. Coordinator C: 10% of contributions

then you could make A be the default preselected coordinator 60% of the times, B 30% of the times etc (by using a random function with probabilities on the client side).

KoalaSat:
Closed in favor of https://github.com/RoboSats/robosats/pull/2315

PR #1633, opened 2024-11-25
Fix book pagination

KoalaSat:
## What does this PR do?
Fix book pagination not working. When clicked, it was always going back to 12

Screenshot from 2024-11-25 11-21-14

## Checklist before merging


KoalaSat:
Tested 👍 merging as is to avoid conflicts

PR #1777, opened 2025-02-23
Update nostr.py

BitcoinVeneto:
## What does this PR do?
This PR fixes an issue where the method add_relays() was renamed to add_relay()

## Checklist before merging


KoalaSat (approved):
Thanks for the quick fix!

PR #1795, opened 2025-03-03
Added Claim Reward Button and improved styling of `Profile.tsx`

### LINKED ISSUE #1776: Create better UX when claiming rewards
**Is your feature request related to a topic?

When users receive rewards the way to claim it is really hidden and dificult to understand.

For security reasons, we don't allow the user to try wirh another invoice if the original fails and we expect them to reach out coordinators, but this is not mentioned at any moment on the UI.

Without changing backend, we need a way to improve this process on the frontend.

**Additional context**
~~Client needs to address all coordinators 1 by 1 and/or the last order (if known). A good stating point is adding a button on the headers of Robots list which opens a new "Claim Reward" popup~~

Once the invoice is sent, would be good to alert users if they didn´t receive the payment, contact its coordinator.

Remember there is also a notification if the clients figures out rewards from the /orders endpoint

UPDATE:

With the new UI, maybe the best option is to add next to KEYS button an additional button CLAIM REWARDS: 

Image

This button will just open any of the coordinator popup with a reward pending for the robot. If there are no rewards available, the button should appear disabled. While the coordinators are loading AND none of the loaded have available rewards, the button should appear loading.

gitsofaryan:
## What does this PR do?
Fixes #1776 

## Description:
- Added a "Claim Reward" button with icon in the profile header for better visibility.
- Created an info alert that explains how to claim rewards by contacting coordinators.
- Displays a list of relevant coordinators to help users know who to contact.

## Checklist before merging

### DIFF
--- a/frontend/src/components/TradeBox/EncryptedChat/index.tsx
+++ b/frontend/src/components/TradeBox/EncryptedChat/index.tsx
@@ -8,7 +8,6 @@ import {
   FederationContext,
   type UseFederationStoreType,
 } from '../../../contexts/FederationContext';
-import { AppContext, UseAppStoreType } from '../../../contexts/AppContext';
 
 interface Props {
   order: Order;
@@ -43,7 +42,6 @@ const EncryptedChat: React.FC<Props> = ({
   status,
 }: Props): React.JSX.Element => {
   const [turtleMode, setTurtleMode] = useState<boolean>(false);
-  const { settings } = useContext<UseAppStoreType>(AppContext);
   const { garage } = useContext<UseGarageStoreType>(GarageContext);
   const { federation } = useContext<UseFederationStoreType>(FederationContext);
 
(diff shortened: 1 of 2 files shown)


gitsofaryan:
@KoalaSat please review my changes.

KoalaSat:
@gitsofaryan sorry for not being more clear on the description. Robots can have a first try to claim their rewards:

Screenshot from 2025-03-05 09-37-00

The idea was to provide this form first (I like your idea of having just one button on the top) and after the user sends its invoice, display the text about contacting coordinators if the payment doesn't arrive.

A couple of extra comments worth checking:

- Because users will be able to claim rewards only when available, you can disable the button until you detect there is one coordinator with rewards for the user (for every coordinator: `robot?.earnedRewards > 0`) 
- I think it's worth to move this reward form to a separate popup that opens with the new button and remove the one on the screenshot above
- I noticed the robot window expands its width when the text appears, careful with that :)
- Don't forget to check for mobile view
- Somehow the robot avatar went out of the margins
Screenshot from 2025-03-05 09-47-04

Thank you so much for this so needed feature :rocket:

KoalaSat:
hi @gitsofaryan , any update here?

PR #1804, opened 2025-03-10
added the alert logic for the coordinators with no on chain swap

### LINKED ISSUE #1786: Alert user for coordinators LN only
Users might realize too late they are choosing/creating an order on a coordinator with no on chain swap option. 

We need to alert somehow before the bond is locked, trying to be less invasive as possible.

khadar1020:
## What does this PR do?
Fixes #1786 
For the coordinators who do not support the on chain payments if the user locked the money in bond and user only supports on chain payments it might be a problem so giving an simple alert message for the user this coordinator does not support on chain payments with out disturbing any flow of his work 

Thank you 

## Checklist before merging

### DIFF
--- a/frontend/src/components/MakerForm/SelectCoordinator.tsx
+++ b/frontend/src/components/MakerForm/SelectCoordinator.tsx
@@ -9,6 +9,7 @@ import {
   type SelectChangeEvent,
   CircularProgress,
   Stack,
+  Alert
 } from '@mui/material';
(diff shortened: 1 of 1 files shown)


KoalaSat:
@khadar1020 the code looks good, but you sure you ran the pre-commit? ideally it should have generated automatically this new sentence on all json locale files

KoalaSat (approved):

PR #1876, opened 2025-04-25
Bugfix: avoid error upon getting succesfully cancelled order

KoalaSat:
## What does this PR do?
Fixes https://github.com/RoboSats/robosats/issues/1245

This PR fixes OrderView to not return error for a successfully cancelled order.

## Checklist before merging


KoalaSat:
Seems like a 204 does not accept a body on the response (which makes sense)

PR #1913, opened 2025-05-09
fix REDIS_URL config instead of environment variable

jerryfletcher21:
## What does this PR do?
REDIS_URL should be read as a config as in [robosats/settings.py](https://github.com/RoboSats/robosats/blob/main/robosats/settings.py)

## Checklist before merging

### DIFF
--- a/robosats/celery/__init__.py
+++ b/robosats/celery/__init__.py
@@ -7,7 +7,7 @@ from celery import Celery
 from celery.schedules import crontab
 
 # You can use rabbitmq instead here.
-BASE_REDIS_URL = os.environ.get("REDIS_URL", "redis://localhost:6379")
+BASE_REDIS_URL = config("REDIS_URL", default="redis://localhost:6379")
 
 # set the default Django settings module for the 'celery' program.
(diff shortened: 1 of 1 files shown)


KoalaSat (approved):

KoalaSat:
Thanks!

KoalaSat:
Seems like ruff complaints about not finding `config`, I tried importing it but didn´t worked https://github.com/RoboSats/robosats/pull/1920 Iĺl revert it to not be  a blocker

jerryfletcher21:
@KoalaSat oh yes sorry I forgot to add that line in the pull request. It is needed thanks.
What do you mean it did not work? The change is needed, otherwise the redis port can not be changed in .env

jerryfletcher21:
@KoalaSat running ruff locally reports no errors/warnings (after adding #1920).
```
ruff check $(git ls-files | grep ".py$")
```

jerryfletcher21:
@KoalaSat I think there is a problem with the github action of ruff, see the comment in #1936. I think this is why you were seeing the error even after you imported `config` in #1920 .

PR #1914, opened 2025-05-09
add option for timing issues in tests

jerryfletcher21:
## What does this PR do?

On my powerful machine, running tests resulted in some errors. After a bit of digging I found out that it was a timing issues.
This PR adds a config option TIMING_EXTRA_IN_TESTS that when set to True, it takes some time where it should so that the tests do not fail.
If the config option is not present (default) nothing is changed.
I had this issue very sporadically on my laptop, and every time when running on an other powerful machine.
I run the tests without docker.

## Checklist before merging

### DIFF
--- a/api/logics.py
+++ b/api/logics.py
@@ -675,6 +675,20 @@ class Logics:
         # Create onchain_payment
         onchain_payment = OnchainPayment.objects.create(receiver=user)
 
+        if \
+            config("TIMING_EXTRA_IN_TESTS", cast=bool, default=False) and \
+            config("TESTING", cast=bool, default=False) and \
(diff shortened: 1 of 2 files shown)


KoalaSat:
Hi @jerryfletcher21 I understand the problem and I think we should find a way to fix it. But I'm not a fan of modifying real code for the sake of testing, I would try to fix the problem from the testing side or at least make usage of the regular tools provider to coordinators.

jerryfletcher21:
Also it is really strange that the github action is failing, because this pull request does not change anything by default, the changes are there just if `TIMING_EXTRA_IN_TESTS` is set.

PR #1928, opened 2025-05-13
Add Nomad payment method

p-manoel:
## What does this PR do?

This PR introduces Nomad as a new payment method to the platform. Nomad is the biggest Brazilian fintech that offers international banking services, allowing users to have a US dollar account and debit card. This addition expands the available payment options for users in Brazil and other regions where Nomad operates.

The integration adds the Nomad icon to our payment methods list, making it available as a selection option for users.

## Checklist before merging


KoalaSat (approved):
thanks!

PR #1931, opened 2025-05-14
allow logging to file and setting LOGGER_LEVEL

jerryfletcher21:
## What does this PR do?
This PR allows logging to a file by specifying `LOG_FILE` in `.env`.
It also allows setting the logging level by specifying `LOGGER_LEVEL` in `.env`.

## Checklist before merging

### DIFF
--- a/.env-sample
+++ b/.env-sample
@@ -42,6 +42,11 @@ LND_GRPC_HOST='localhost:10009'
 
 REDIS_URL='redis://localhost:6379/1'
 
+# If set, log to file specified
+# LOG_FILE="<path>"
+# Change logger level (default is "WARNING")
+# LOGGER_LEVEL="WARNING"
+
 # List of market price public APIs. If the currency is available in more than 1 API, will use median price.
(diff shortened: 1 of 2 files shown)


.env-sample
```diff
@@ -42,6 +42,13 @@ LND_GRPC_HOST='localhost:10009'
 
 REDIS_URL='redis://localhost:6379/1'
 
+# If set to True, load to console (default is False)
+# LOG_TO_CONSOLE=False
```
KoalaSat, line 46: Most coordinators uses https://github.com/RoboSats/robosats-deploy and there LOG_TO_CONSOLE is set to 1. Can you change the default to True?

jerryfletcher21: Do you want to change it just in the `.env-sample`, or also in the code so as to make `True` the default?
Because before it was not the default

jerryfletcher21: yes I did not want to change any logic. Change done

KoalaSat (approved):

PR #1936, opened 2025-05-16
fix ruff warning in telegram_watcher

jerryfletcher21:
## What does this PR do?
I keep on forgetting to run ruff locally.
There is a problem with the github actions though. Because ruff starts complaining after a pull request has been merged, so it seems that the following pull requests are giving problems when in fact it was the pull request before (as for example in #1920 and in this one, that is fixing the error reported by ruff).

## Checklist before merging


KoalaSat (approved):

KoalaSat (commented):
Hi @jerryfletcher21 , looks like ruff doesn´t like this change: https://github.com/RoboSats/robosats/actions/runs/15073095693/job/42374088387?pr=1936

PR #1969, opened 2025-05-28
change lnd down errors to general node errors

jerryfletcher21:
## What does this PR do?
This PR changes the error messages that mentions LND to be general error messages about the lightning node, since the backend can be LND or core lightning

## Checklist before merging


KoalaSat (approved):
Nice!

PR #1985, opened 2025-06-02
ensure nostr pubkey is present in auth when creating robot

jerryfletcher21:
## What does this PR do?
Since nostr is becoming necessary for the trade process, this PR checks that the nostr_pubkey is present in the authorization header when creating a robot.
It also documents the presence of nostr_pubkey in some comments.

## Checklist before merging

### DIFF
--- a/robosats/middleware.py
+++ b/robosats/middleware.py
@@ -30,11 +30,12 @@ class DisableCSRFMiddleware(object):
 class SplitAuthorizationHeaderMiddleware(MiddlewareMixin):
     """
     This middleware splits the HTTP_AUTHORIZATION, leaves on it only the `Token ` and creates
-    two new META headers for both PGP keys.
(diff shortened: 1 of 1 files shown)


KoalaSat:
This change can't be possible while some folks run old frontend versions that are not sending nostr pubkey. Fortunately for us robodats clients always force users to upgrade if they detect a major new version on any of the coordinatros. That means these changes will  only be possible with v0.8.0.

jerryfletcher21:
@KoalaSat yes I understand. But it has been reported by other coordinators that there are now problems in sending the messages and completing the orders when nostr_pubkey is not present. It think it may be better to block the creation of robots all together and force them to update instead of allowing to create robots and not being able to complete the trades for old clients.

KoalaSat (approved):

PR #2003, opened 2025-06-16
documentation: update Wallets file

htteot:
## What does this PR do?
Fixes #<ISSUE_NUMBER>

## Checklist before merging

### DIFF
--- a/api/logics.py
+++ b/api/logics.py
@@ -1308,9 +1308,9 @@ class Logics:
         bond_satoshis = int(order.last_satoshis * order.bond_size / 100)
 
         if user.robot.wants_stealth:
-            description = f"{config("NODE_ALIAS")} - Payment reference: {order.reference}. This payment WILL FREEZE IN YOUR WALLET, check on RoboSats if the lock was successful. It will be unlocked (fail) unless you cheat or cancel unilaterally."
+            description = f"{config("NODE_ALIAS")} - Payment reference: {order.reference}. This payment WILL FREEZE IN YOUR WALLET, check on the trading platform if the lock was successful. It will be unlocked (fail) unless you cheat or cancel unilaterally."
         else:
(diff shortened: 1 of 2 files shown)


KoalaSat:
Hi! Can you expose here the reasons to remove those wallets? Just to keep track of it

htteot:
Wallets Mutiny and Phoenix have been removed due to:

- Mutiny:
-- wallet shutdown and the end of support:
--- https://blog.mutinywallet.com/mutiny-wallet-is-shutting-down/

- Phoenix:
-- wallet team does not want to support the lock invoice used by Robosats:
--- chat conversation: "Hello, Locking payments is a hack that puts strain on the network. Phoenix won't allow unreasonably high lock time. But if the lock time is reasonable, then it should work.  Best,  Phoenix support" 
---- Based on the tests, all transactions using this wallet failed in Robosats so I am removing this wallet. Waiting for their policy to change before adding it back.

KoalaSat:
> Wallets Mutiny and Phoenix have been removed due to:
> 
> * Mutiny:
>   -- wallet shutdown and the end of support:
>   --- https://blog.mutinywallet.com/mutiny-wallet-is-shutting-down/
> * Phoenix:
>   -- wallet team does not want to support the lock invoice used by Robosats:
>   --- chat conversation: "Hello, Locking payments is a hack that puts strain on the network. Phoenix won't allow unreasonably high lock time. But if the lock time is reasonable, then it should work.  Best,  Phoenix support"
>   ---- Based on the tests, all transactions using this wallet failed in Robosats so I am removing this wallet. Waiting for their policy to change before adding it back.

I see now. I agree on removing Mutiny if the wallet is shutting down. But that's not the case of Phoenix so people will continue trying to use it for Robosats, I think that we should just update the row instead or removing it. What do you think?

htteot:
I agree

htteot:
I have also updated logics.py file to replace string 'RoboSats' with string 'the trading platform'.
This is to increase the privacy and make surveillance a bit less easy.
Currently every transaction contains a reference to RoboSats platform exposing what the user of the Lightning wallet is doing.
This data can be potentially collected by the Lightning wallet operator or Lightning liquidity providers.

KoalaSat (approved):

PR #2024, opened 2025-06-26
middleware cleaning

jerryfletcher21:
## What does this PR do?
This PR removes some old code from `middleware.py`.
Since #1985 was merged, it removes the legacy support for splitting `HTTP_AUTHORIZATION` in 3 instead of 4.
It removes support for sending authorization with cookies that really old clients pre federation were using.
I am pretty sure nobody is running such an old client that still sends cookies, but with #1985 merged these will no longer work since they were not sending the nostr pubkey.

## Checklist before merging

### DIFF
--- a/robosats/middleware.py
+++ b/robosats/middleware.py
@@ -42,12 +42,7 @@ class SplitAuthorizationHeaderMiddleware(MiddlewareMixin):
         auth_header = request.META.get("HTTP_AUTHORIZATION", "")
         split_auth = auth_header.split(" | ")
 
-        if len(split_auth) == 3:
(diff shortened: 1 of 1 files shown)


jerryfletcher21:
@KoalaSat with #1985 merged, the old clients that did not send the nostr pubkey can no longer create robots. This PR simply cleans up old code from clients that did not send the nostr pubkey. Since #1985 was merged, it should not be controversial I think.

jerryfletcher21:
@KoalaSat I can help with the transition of the chat to nostr in the backend

PR #2155, opened 2025-08-09
Escape characters using two backslashes

aftermath2:
## What does this PR do?

This PR restores the clipboard content validation and updates the expression characters to escape characters using two backslashes instead of one. Java is fine using a single backslash but android studio fails to compile.

Related to https://github.com/RoboSats/robosats/pull/2153.

## Checklist before merging

### DIFF
--- a/android/app/src/main/java/com/robosats/WebAppInterface.kt
+++ b/android/app/src/main/java/com/robosats/WebAppInterface.kt
@@ -41,7 +41,7 @@ class WebAppInterface(private val context: MainActivity, private val webView: We
 
     // Security patterns for input validation
     private val UUID_PATTERN = Pattern.compile("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", Pattern.CASE_INSENSITIVE)
-    private val SAFE_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9\\s_\\-.,:;!?()\\[\\]{}\\"]*$")
+    private val SAFE_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9\\s_\\-.,:;!?()\\[\\]{}\"]*$")
 
     // Maximum length for input strings
     private val MAX_INPUT_LENGTH = 1000
--- a/android/app/src/main/java/com/robosats/WebAppInterface.kt
+++ b/android/app/src/main/java/com/robosats/WebAppInterface.kt
@@ -41,7 +41,7 @@ class WebAppInterface(private val context: MainActivity, private val webView: We
 
     // Security patterns for input validation
     private val UUID_PATTERN = Pattern.compile("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", Pattern.CASE_INSENSITIVE)
-    private val SAFE_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9\s_\-.,:;!?()\[\]{}\"]*$")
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/robosats/WebAppInterface.kt
```diff
@@ -41,7 +41,7 @@ class WebAppInterface(private val context: MainActivity, private val webView: We
 
     // Security patterns for input validation
     private val UUID_PATTERN = Pattern.compile("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", Pattern.CASE_INSENSITIVE)
-    private val SAFE_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9\s_\-.,:;!?()\[\]{}\"]*$")
+    private val SAFE_STRING_PATTERN = Pattern.compile("^[a-zA-Z0-9\\s_\\-.,:;!?()\\[\\]{}\\"]*$")
```
KoalaSat, line 44: Android Studio is still warning on this one

KoalaSat: I have been reviewing in deep and I'm pretty confident the real reason of the reported bug was this: https://github.com/RoboSats/robosats/pull/2154/files#diff-5726100c3de4fda01c1c98e342f00e99339efca15c82e26416c06e70e7f6fef7L114

aftermath2: Before it was exactly the same except for the last double quote character (`\\"`). If you remove that it compiles?

I think it might be causing the string to end earlier and cause the syntax error. It is a pity I cannot test it though, sorry

aftermath2: > I have been reviewing in deep and I'm pretty confident the real reason of the reported bug was this

Yes, but it entered that if block because the regex did not match I think

aftermath2: I mean removing the validation entirely is an option but you lose that security check

aftermath2: Just tested it here https://pl.kotl.in/wg7ssNcfQ. It was because the double quote actually has to be escaped with only one back slash 🤪

KoalaSat: Yeah now that format works. Let's do one thing, let's recover the validation but remove from `isValidInput` the length check, I though it was a good idea but later I realized that chat logs can be extremely long. I see no point on that validation.

aftermath2: Sure!

KoalaSat (approved):
Thank you!

Please accept a tip for your work. Paste here a LN invoice for 40,000 sats with a long exploration date.

aftermath2:
@KoalaSat Thanks!

```
lnbc400u1p5fwhswsp5qtamzpmx2g7p4jf6vzehqatcpdknn6dwdv3742vcrxxazzn3ru5qpp56hgfwnfqegxefkuje4pgx6ch34696zn5fezachhnszpn8p5guq5qhp5uwcvgs5clswpfxhm7nyfjmaeysn6us0yvjdexn9yjkv3k7zjhp2sxq9z0rgqcqpnrzjqtc63jrkql6ptj8j9sq9jvqzwav5rh4y3p5uugcfdte8kr8aes9kjzufjuqqwxqqqqqqqqqqqqqqqeqqjq9qxpqysgq04pfug6g38ms2srfcr0thgu3a37r0vjejsm6kxtae88n7rv7dxszuatwalrwpjz0krrycfslpjss32lyk9pj0p3ehzgkx8udchc5z4cqrhvzx0
```

RobosatsDevFund:
`9cde72450bdb56135edd5b99568874a4c5d4c0021cd0c0a48aca800e367b183a`

PR #2165, opened 2025-08-12
Fix payment method icons

aftermath2:
## What does this PR do?

This PR fixes a bug in the payment method icons matching.

Payment method icons were matched using exact casing matches. 

https://github.com/RoboSats/robosats/pull/2131 fixed it by converting to lowercase but introduced a bug in the order details page (which I didn't test, my bad) by displaying the icon and the method name.

The change was reverted in https://github.com/RoboSats/robosats/pull/2159. Fixing the latter issue but introducing again the former.

Here I modify the matching again but instead replace the method name as it is from the custom methods instead of replacing the lowercase version of it, effectively getting rid of both bugs.

### Tests

<details><summary>Before</summary>

premium

</details>

<details><summary>After</summary>

premium_3

</details>

## Checklist before merging

### DIFF
--- a/frontend/src/components/PaymentMethods/StringAsIcons.tsx
+++ b/frontend/src/components/PaymentMethods/StringAsIcons.tsx
@@ -33,10 +33,10 @@ const StringAsIcons: React.FC<StringAsIconsProps> = ({
 
     // Adds icons for each PaymentMethod that matches
     methods.forEach((method, i) => {
-      const regex = new RegExp(`\\b${method.name}\\b`, 'g');
+      const regex = new RegExp(`\\b${method.name.toLowerCase()}\\b`, 'g');
 
-      if (regex.test(customMethods)) {
-        customMethods = customMethods.replace(regex, '');
+      if (regex.test(customMethods.toLowerCase())) {
+        customMethods = customMethods.replace(method.name, '');
         rows.push(
           <Tooltip
             key={`${method.name}-${i}`}


frontend/src/components/PaymentMethods/StringAsIcons.tsx
```diff
@@ -33,10 +33,10 @@ const StringAsIcons: React.FC<StringAsIconsProps> = ({
 
     // Adds icons for each PaymentMethod that matches
     methods.forEach((method, i) => {
-      const regex = new RegExp(`\\b${method.name}\\b`, 'g');
+      const regex = new RegExp(`\\b${method.name.toLowerCase()}\\b`, 'g');
 
-      if (regex.test(customMethods)) {
-        customMethods = customMethods.replace(regex, '');
+      if (regex.test(customMethods.toLowerCase())) {
+        customMethods = customMethods.replace(method.name, '');
```
aftermath2, line 39: Here is the missing part from the original PR, replace the method name as is to remove it from the list instead of replacing the lowercase version of it

KoalaSat: The problem why I removed this is that you can receive a manually added `pix` instead of `PIX`, match the regexp, add the icon but yet keep thew original `pix` as custom method

aftermath2:
@KoalaSat Thank you!

```
lnbc300u1p5fcuxmsp5kujz8df5pmvejetxsjna9e4tq4zfeyj4gk988vvv0gr759tafa9spp5d5xu80ahyr6r5e7w9fqmvn8xj43maa48erd22fcmchmn7x5jr8jqhp5uwcvgs5clswpfxhm7nyfjmaeysn6us0yvjdexn9yjkv3k7zjhp2sxq9z0rgqcqpnrzjqw7c9dkkx4nur9sztw2zzpzj8u8rgsqgdsykylg5pwplh26824lc7rvlqcqqn3gqqyqqqqlgqqqqqqgq2q9qxpqysgqnnhne3jtwfmw7dunyzp0h6v0lyqnszgxxvrpv2nurczu4wewn8rj7ez2j87mju3edyxrgxha5my7zx3xn2djvyfru2uyxgl9xmem2tsqeq9vz4
```

RobosatsDevFund:
`4c51e8f01ab23fb91564770c17b74ca1cb07e9d6ab784045e174fdb74ffea585`

PR #2312, opened 2025-12-04
update alpine for SOCKS5 support

MattDHill:
## What does this PR do?
Updates Alpine to 3.23. SOCKS5 support was introduced in 3.19.

This makes it better for StartOS runtime.

## Checklist before merging


KoalaSat:
Thank you @MattDHill !

PR #2344, opened 2026-01-06
fix: correct exchange rate calculation for taken range orders

### LINKED ISSUE #2336: bug: order summary: incorrect "You receive" amount
**Describe the bug**
The "Your receive" satoshi amount when buying bitcoin as maker shown in the "Order" tab of the ongoing trade during the chat period is incorrect. I noticed in multiple trades that the amount shown is much lower than the actual trade amount. 

**To Reproduce**
1. Make offer to buy bitcoin for Euro in an amount range on Temple of Sats
2. Wait for it to get taken
3. Once in the chat, open the "Order" tab
4. See too low "You receive" amount.

**Expected behavior**
"You receive" should show the amount for which the user submitted the lightning payout invoice in the previous step.

**Screenshots**
(should be >1m sat here at the current exchange rate):
Image

 - OS: Android, Browser, independent of system
 - Version: 0.8.3

keshav0479:
Fixes #2336

### Problem
For range orders, the "You receive" satoshi amount in the Order tab showed an incorrect (lower) value after the order was taken.

### Cause
The rate calculation used `max_amount` whenever `has_range` was true, even after the order was taken and `amount` was set to the actual taken value. This caused a math mismatch:
- `satoshis_now` was computed for the taken amount (eg 800 EUR)
- `rate` was computed incorrectly using `max_amount` (eg 1000 EUR)

### Fix
Changed the rate calculation to prioritize `order.amount` (the actual taken amount) when available. It falls back to `max_amount` only for untaken range orders.

### DIFF
--- a/frontend/src/components/OrderDetails/index.tsx
+++ b/frontend/src/components/OrderDetails/index.tsx
@@ -196,7 +196,7 @@ const OrderDetails = ({
     const defaultRoutingBudget = 0.001;
     const btc_now = order.satoshis_now / 100000000;
     const rate =
-      (order.has_range && order.max_amount ? Number(order.max_amount) : Number(order.amount)) /
+      (order.amount && order.amount > 0 ? Number(order.amount) : Number(order.max_amount)) /
       btc_now;
 
     if (isBuyer) {


keshav0479:
Thank you @KoalaSat! I really appreciate the review and the tip.

lnbc350u1p54e7happ5pxjgfjehf2uw3kj0p8x3cudyk3pyf9w69tsdyd2stfesx8h6p2xqcqzyssp5djelhlapyhgdre040ulwq7raml9efm5xauv9l6n8s9hp6nxjpqrq9q7sqqqqqqqqqqqqqqqqqqqsqqqqqysgqdqqmqz9gxqyjw5qrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glcll63a0scqz7stsqqqqlgqqqqqeqqjqxvjgjpcq6hrwwj9ugyjpavukl7sun0pfut4ghzcwku60264ldepx2tss8v4a4yv7jnfnpn87v2y8m9wesdtj6kngcrgqkepzakcfc8cq49008x

RobosatsDevFund:
@keshav0479 Seems like it is not possible to pay this invoice due to either your channels are full and your node is and Robosats' node is poorly connected. You can try with a proxy wallet.

keshav0479:
> @keshav0479 Seems like it is not possible to pay this invoice due to either your channels are full and your node is and Robosats' node is poorly connected. You can try with a proxy wallet.




`lnbc350u1p54lqtkpp5p8lgw566cpujcf2umd002jvyw3psetrms53fmj3fss79a9h2xy5sdqqcqzzsxqyz5vqsp5qvda07ev5qavmrlj2we5w75p7etfaw3l08vp6axp452pu6j3227q9qxpqysgqcx8l6sd6kwlvn7d2v9pjtsgma5gfvgctr0847aqgjdg4xeyglx95wvj5jetzz5palzg6gmtht9nj8g8av9nxdqxjdlk74srxppe0m4speefc5n`

RobosatsDevFund:
7390939451b5ae8f9942040ba440325e463d1656553f327f530908e81be098ab

PR #2363, opened 2026-01-12
feat: Implement Garage Key functionality and account recovery

### LINKED ISSUE #2342: Garage Key
The full implementation of this issue will be rewarded with ⚡ **800,000 sats** ⚡

## Is your feature request related to a problem? Please describe.

If users want to maintain their privacy, they are required to constantly change their robot keys for each trade and manage a secure storage system for them. This method is clearly inefficient and leads some users to reuse their keys, resulting in a loss of privacy.

This issue aims to implement a method that allows users to store and use a single key while enabling them to utilize a different robot key for each trade.

For users, we will completely remove the concept of "Robot token" in favor of a new "Garage Key," which is based on a bech32 encoded entity defined in BIP-32 or in NIP-19 (https://github.com/nostr-protocol/nips/blob/master/19.md).

### Encoding/Decoding

The garage key will use the bech32 prefix `robo` and will store the key itself. To ensure entropy, it will reuse the same generation method used for robot tokens but will result in a bech32 string instead of a bech58. 

Example: The key `3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d` translates to `robo180cvv07tjdrrgpa0j7j7tmnyl2yr6yr7l8j4s3evf6u64th6gkwsyjh6w6`.

### Derivation Path

Bitcoin uses the derivation path `m/44'/0'/0'/0`.

```
m
│
├── 44'
│   ├── 0' (Bitcoin)
│   │   ├── 0' (Account 0)
│   │   │   ├── 0 (External)
│   │   │   └── 1 (Internal)
```

We can define our own, and a potentially suitable one would be `m/44'/88'/n'/0`. From the garage view, users will be able to change the Account value n of the derivation path to access a brand new robot deterministically derived from its Garage key.

### Latest Account

If the user recovers their Garage Key, the app should determine the most recently used account. However, querying N number of accounts from M number of Coordinators poses significant scaling issues. To address this, we can leverage the federation’s relays to store garage metadata that can be fetched and updated at any time.

A nostr secret key will be calculated from the decoded plain key of the Bech32 Address (`sha256Hash(sha512(key));`). This nostr key will be used to publish and fetch an app configuration event (https://github.com/nostr-protocol/nips/blob/master/78.md) inside of a gift wrapped event (https://github.com/nostr-protocol/nips/blob/master/59.md) sent to itslef :

```
{
    created_at: Math.floor(Date.now() / 1000),
    kind: 30078,
    tags: [
        ['account', '3']
    ],
    content: '',
    pubkey: '123...789'
}
```

The example above defines a garage where the latest account used to create an order is 3 (`m/44'/88'/3'/0`).

This event will be published to federation relays every time a new robot creates a new order on a Coordinator. When the app loads with a Garage Key, it will attempt to fetch this event and use it to display the most recent account number. If nothing is found after receiving `EOSE` from all connected relays, the garage will fall back to Account 0.

### Steps

The Garage Key will be used to generate an (in practice) unlimited number of bech58 robot tokens. These robot tokens will remain out of users' sight but will be used as usual in the background.

1. For a given Bech32 garage key (`robo1...yjh6w6`)
2. Decode it to obtain the plain key (`3bf...59d`)
3. Calculate its nostr pubkey (`'123...789'`)
4. Find in relays the most recent account number (ex: `18`) or fall back to `0`  to determine the derivation path (`m/44'/88'/18'/0` or `m/44'/88'/0'/0`)
5. Apply the derivation path to the plain key to obtain the derived key
6. Encode the derived key to Bech58 to get the robot token to be used and displayed in Robot view.

### Requirements

1. Implement all the encryption logic to be used as defined above.
2. Create a radio selector to switch between the old robot token view and the new garage key view to ensure a smooth transition. Probably worth it to not modify `RobotPage` files and just duplicate and modify them to a new `GaragePage` folder. Both views must end up on defining what's the active robot. Default view must be `GaragePage` .
3. Update `GaragePage/Onboarding.tsx` to explain how this new process works
4. Change the robot view to easily allow users to change Account numbers for the same Garage key, as replacement of the actual robot drop down. "Add" and "Delete" robot buttons should be replaced by next and previous.
6. Multiple garage key management should NOT be implemented.
7. Implement latest Account recovery process including fetching and publication.
 
### Tech details 

- No changes should be implemented in the API side
- Multiple garage `Slots` doesn't makes sense anymore. We don't require to store and manage in parallel all robot keys (maybe rename 'Slot' to 'Account'?) . This way, only one single Slot attribute can be left in Garage and updated every time the user changes the Account number. Remember to not remove the old code and keep it consistent to avoid breaking changes.
- Frontend should not allow used robots to create new orders.
- This issue can be implemented in multiple PRs to ensure a good feedback loop and help  reviewers. All merged to the same feature branch: https://github.com/RoboSats/robosats/tree/garage-key

eudaldgr:
## What does this PR do?
Fixes #2342 

This PR implements Garage Key support and account recovery, introducing a new key-based mechanism for managing robot accounts and recovery flows.

## Summary of changes:
- Added GarageKey model to manage garage keys, including generation, encoding, decoding, and account index management.
- Enhanced Garage model to support garage key mode, including methods for creating robots from garage keys and managing account indices.
- Updated RecoveryDialog to validate and handle garage keys for robot recovery.
- Introduced account recovery events for managing robot accounts through relays.
- Modified StoreTokenDialog to differentiate between storing garage keys and robot tokens.
- Implemented settings for switching between legacy and garage key modes, including user prompts for data safety.
- Enhanced Slot model to determine if a slot is reusable based on the last order status.
- Updated TradeBox to handle bad requests and provide user feedback on order creation.
- Added utility functions for garage key operations, including validation and token derivation.

## Checklist before merging

### DIFF
--- a/frontend/src/basic/GaragePage/AccountNavigator.tsx
+++ b/frontend/src/basic/GaragePage/AccountNavigator.tsx
@@ -1,0 +1,82 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import {
+  Box,
+  IconButton,
+  Typography,
+  Tooltip,
+  CircularProgress,
+} from '@mui/material';
+import { ChevronLeft, ChevronRight } from '@mui/icons-material';
+
+interface AccountNavigatorProps {
+  accountIndex: number;
+  onPrevious: () => void;
+  onNext: () => void;
+  loading?: boolean;
+  disabled?: boolean;
+}
+
+const AccountNavigator = ({
+  accountIndex,
+  onPrevious,
+  onNext,
+  loading = false,
+  disabled = false,
+}: AccountNavigatorProps): React.JSX.Element => {
+  const { t } = useTranslation();
+
+  return (
+    <Box
+      sx={{
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        gap: 1,
+      }}
+    >
+      <Tooltip title={t('Previous account')} placement='top'>
+        <span>
+          <IconButton
+            onClick={onPrevious}
+            disabled={disabled || loading || accountIndex === 0}
+            size='small'
+          >
+            <ChevronLeft />
+          </IconButton>
+        </span>
+      </Tooltip>
+
+      <Box
+        sx={{
+          display: 'flex',
+          alignItems: 'center',
+          minWidth: '6em',
+          justifyContent: 'center',
+        }}
+      >
+        {loading ? (
+          <CircularProgress size={20} />
+        ) : (
+          <Typography variant='body1' sx={{ fontFamily: 'monospace' }}>
+            {t('Account')} #{accountIndex}
+          </Typography>
+        )}
+      </Box>
+
+      <Tooltip title={t('Next account')} placement='top'>
+        <span>
+          <IconButton
+            onClick={onNext}
+            disabled={disabled || loading}
+            size='small'
+          >
+            <ChevronRight />
+          </IconButton>
+        </span>
+      </Tooltip>
+    </Box>
+  );
+};
+
+export default AccountNavigator;
--- a/frontend/src/basic/GaragePage/GarageKeyInput.tsx
+++ b/frontend/src/basic/GaragePage/GarageKeyInput.tsx
@@ -1,0 +1,117 @@
+import React, { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import {
+  IconButton,
+  TextField,
+  Tooltip,
+  InputAdornment,
+  CircularProgress,
+} from '@mui/material';
+import { ContentCopy, Check } from '@mui/icons-material';
+import { validateGarageKey } from '../../utils';
+
+interface GarageKeyInputProps {
+  garageKey: string;
+  setGarageKey: (key: string) => void;
+  editable?: boolean;
+  loading?: boolean;
+  onPressEnter?: () => void;
+  autoFocusTarget?: 'textField' | 'copyButton';
+  label?: string;
+}
+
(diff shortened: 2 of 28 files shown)


KoalaSat:
I have this PR https://github.com/RoboSats/robosats/pull/2358 in queue waiting after next release to be merged, and I believe is going to kind of affect this one, can you rebase that branch or wait to be merged to main? 

I believe the only conflict you'll have is that I got rid of multiple 1059 nostr kind subscriptions and centralized everything in https://github.com/RoboSats/robosats/pull/2358/files#diff-5d4e28e67be411858c9a44e3954fcd72636fdf56e3c83b38376740495d89b3d7R71 so you can fetch https://github.com/RoboSats/robosats/pull/2358/files#diff-5d4e28e67be411858c9a44e3954fcd72636fdf56e3c83b38376740495d89b3d7R53

eudaldgr:
Got it, thanks for the context. I’ll review PR #2358 now and make the necessary changes on my side.

eudaldgr:
@KoalaSat you closed it, the changes are no longer needed or do you have to open it again because it was closed automatically when applying the other PR?

KoalaSat:
@eudaldgr I think it just happened automatically after merging my PR, the one I asked you to rebase. Not clear for me why, can you re-open and make sure it's now re-based with `main` ?

eudaldgr:
@KoalaSat i can't reopen, can you?

KoalaSat:
Nope, can you try to open anew PR? Seems like the reason is that my branch has been automatically deleted by GitHub and it affected yours

PR #2038, opened 2025-07-02
add option for routing_budget_ppm in /api/rewards

jerryfletcher21:
## What does this PR do?
This PR uniforms the `/api/rewards` with the `/api/order` `update_invoice` (changed in #326), adding the option `routing_budget_ppm` in `/api/rewards`, and thus fixing #1382 for the backend.
An other pull request implementing it in the frontend is needed (maybe to be incorporated in #1973). This commit still leaves the option to not specify `routing_budget_ppm` in order to be compatible with old clients, with the notice in the code that it should be deprecated in the future.
It also remove the `MIN_FLAT_ROUTING_FEE_LIMIT` variable from `.env-sample`, since after #347 it is no longer used.

## Checklist before merging
- [x] Install [pre-commit](https://pre-commit.com/#installation) and initialize it: `pip install pre-commit`, then `pre-commit install`. Pre-commit installs git hooks that automatically check the codebase. If pre-commit fails when you commit your changes, please fix the problems it points out.

PR #2364, opened 2026-01-13
feat:add routing budget ui to rewards claim

### LINKED ISSUE #1382: Claiming rewards does not support routing_budget_ppm
**Describe the bug**
The `/api/order` `update_invoice` and `/api/rewards` requests are similar: they accept a signed invoice to be paid by the coordinator. The amounts of the invoice are also similar, since trades ended up in a dispute will be refunded by rewards.
The first one supports the parameter `routing_budget_ppm` that allows the user to specify the routing budget (to be subtracted from the invoice amount), while the second one does not support it and instead uses the default of 1000 ppm or 10 sats.

in [api/logics.py](https://github.com/RoboSats/robosats/blob/main/api/logics.py):
update_invoice:
```
routing_budget_sats = float(num_satoshis) * (
    float(routing_budget_ppm) / 1_000_000
)
num_satoshis = int(num_satoshis - routing_budget_sats)
```
withdraw_rewards:
```
num_satoshis = user.robot.earned_rewards
routing_budget_sats = int(
    max(
        num_satoshis * float(config("PROPORTIONAL_ROUTING_FEE_LIMIT")),
        float(config("MIN_FLAT_ROUTING_FEE_LIMIT_REWARD")),
    )
)  # 1000 ppm or 10 sats
```

**Expected behavior**
I think also `/api/rewards` should support `routing_budget_ppm`.
The drawback is that by supporting it, the invoice amount will change depending on the value of `routing_budget_ppm`, so in the main client there will have to be a similar page like for `update_invoice` that let changing the value of `routing_budget_ppm` and showing the resulting invoice amount.

keshav0479:
## What does this PR do?
Fixes #1382

This PR implements the frontend UI for setting `routing_budget_ppm` when claiming rewards.

### Changes
- Added a `Routing Budget (PPM)` input field to the Claim Rewards dialog.
- Updated `Robot.model.ts` to pass the `routing_budget_ppm` parameter to the backend API.
- Users can now specify a routing budget (default 1000 PPM) instead of relying on the backend default.
Screenshot From 2026-01-13 19-27-05

### Context
This completes the feature introduced in #2038 (which added backend support) by providing the necessary UI in the frontend.

### Verification
- Tested locally(mock): The input field appears correctly and passes the value to the API.

## Checklist before merging


KoalaSat:
@keshav0479 can you do something with that "Claim" button, it already looked ugly before, it would be great if we make it look nicer now that  we are touching

keshav0479:
you are totally right, it does need some love, i'll make sure it looks crisp and aligns better with the  UI. Working on it

keshav0479:
Screenshot From 2026-01-16 00-09-36
Screenshot From 2026-01-16 00-09-51
Screenshot From 2026-01-16 00-18-37

How about this ? Let me know if you'd like any tweaks.

KoalaSat:
@keshav0479 awesome stuff

KoalaSat:
@keshav0479 seems like there are some 500 codes from the backend, we are checking, but this PR looks great :+1: in the mid time, do you think it's possible to keep the trophy icon after displaying the form? just to keep consistency. 

I'll merge after we figure out what's wrong in backend.

keshav0479:
Done! trophy icon now stays visible in the expanded form 
Screenshot From 2026-01-16 19-10-59

keshav0479:
> @keshav0479 seems like there are some 500 codes from the backend, we are checking, but this PR looks great 👍 in the mid time, do you think it's possible to keep the trophy icon after displaying the form? just to keep consistency.
> 
> I'll merge after we figure out what's wrong in backend.

Regarding the 500 errors - I reviewed the code flow and confirmed my changes are purely UI/layout, so the API calls remain unchanged also the `routing_budget_ppm` was already supported since #2038. Happy to help dig into backend logs if you'd like an extra pair of eyes

keshav0479:
@KoalaSat coming back to this one after a while.

While rebasing it locally i checked reward backend path again too. Not sure if this is the same 500 you saw without logs, but i did find one real issue in same `/api/reward/` flow: LND payout path could return plain `False`, while `withdraw_rewards()` expects `(paid, failure_reason)`

Also noticed custom routing budget from #2038 was not getting stored on reward `LNPayment`, so the actual payout still ended up using default fee cap.

Opened a PR for that backend fix first. Once that part looks good, i’ll rebase this PR on top of it.

PR #2368, opened 2026-01-13
fix:return 200 OK on successful order cancel

### LINKED ISSUE #1245: POST api/order/ cancel returns error 400 bad_request when successful
**Describe the bug**
Other POST api/order/ `action` do not return error when successful, but `cancel` returns error 400 "bad_request": "This order has been cancelled by the maker". I think this is because when requesting an order that is canceled this is what is returned, but while doing the cancelling action is weird to be returned an error when successful. 
Maybe to fix, when requesting an order that is cancelled, it can be returned a json without error, something like "status": "This order has been cancelled by the maker".

in api/views.py:
```
        # 2) If order has been cancelled
        if order.status == Order.Status.UCA:
            return Response(
                {"bad_request": "This order has been cancelled by the maker"},
                status.HTTP_400_BAD_REQUEST,
            )
        if order.status == Order.Status.CCA:
            return Response(
                {"bad_request": "This order has been cancelled collaborativelly"},
                status.HTTP_400_BAD_REQUEST,
            )
```

**To Reproduce**
Make a POST request at api/order/ with the `action` `cancel`.

**Expected behavior**
Not to be returned an error when then cancel is successful.

keshav0479:
## What does this PR do?
Fixes #1245

### What was happening?
When a user successfully cancelled their order, the API was returning a confusing `400 Bad Request` error — even though the cancellation actually worked! This happened because after cancelling, the code fell through to `self.get()`, which sees "oh, this order is cancelled" and returns an error.

### What this PR does
Simple fix: After a successful cancellation, we now return `200 OK` immediately with the updated order info (`id`and `status`), so the frontend knows everything went smoothly.
### Why return `{id, status}` instead of a full response?
The frontend uses `Object.assign()` to merge updates, so we only need to send what changed. This keeps things minimal and plays nicely with the existing frontend logic.

### Testing
- Reviewed frontend `Order.update()` — confirms it handles partial updates correctly
- Code analysis verified the fix prevents the fall-through bug
Thanks @KoalaSat for the guidance on keeping the frontend in mind!

## Checklist before merging

### DIFF
--- a/tests/test_trade_pipeline.py
+++ b/tests/test_trade_pipeline.py
@@ -1110,6 +1110,7 @@ class TradeTest(BaseAPITestCase):
 
         self.assertEqual(data["id"], trade.order_id)
         self.assertEqual(data["status"], Order.Status.UCA)
+        self.assertEqual(data["bad_request"], "This order has been cancelled")
 
         maker_headers = trade.get_robot_auth(trade.maker_index)
         maker_nick = read_file(f"tests/robots/{trade.maker_index}/nickname")
@@ -1220,6 +1221,7 @@ class TradeTest(BaseAPITestCase):
         self.assertEqual(trade.response.status_code, 200)
         self.assertEqual(data["id"], trade.order_id)
         self.assertEqual(data["status"], Order.Status.UCA)
+        self.assertEqual(data["bad_request"], "This order has been cancelled")
 
         trade.get_order(trade.taker_index)
         data = trade.response.json()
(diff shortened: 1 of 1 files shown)


fxrstor (approved):

KoalaSat:
Ey @keshav0479 thank you for working on this. Can you please add some checks for this status in `tests/test_trade_pipeline.py` ?

keshav0479:
Sure , i'll add the checks for the cancel status in `tests/test_trade_pipeline.py` and update the PR shortly. Thanks for the validation

keshav0479:
Done,i've updated the PR with checks for the cancel status in  `tests/test_trade_pipeline.py` also refined the fix slightly to ensuring we only return the partial 200 OK response for actual cancellations (UCA/CCA). For other cancel like actions (like a pretaker canceling), it still falls through to return the full order object, preserving existing behavior for those edge cases.

keshav0479:
Thanks for guidance throughout @KoalaSat ! 
```
```

KoalaSat:
@robosatsdevfund

RobosatsDevFund:
a870345efbb3c27a66e0cbf912d6e188fd33d9e3198a76709f6c2bb440307f8c

PR #2377, opened 2026-01-16
feat: make nostr host configurable

alicecoordinator:
## What does this PR do?
Makes it possible for coordinators to configure the nostr relay host.

## Checklist before merging


jerryfletcher21:
Nice!
Are you using the traditional scripts for your coordinator?
There is one other instance where you should add it, in [traditional/templates/strfry.conf](https://github.com/RoboSats/robosats/tree/main/scripts/traditional/templates/strfry.conf) under `relay` -> `bind`.
And then adding the `sed` command in [traditional/traditional-services](https://github.com/RoboSats/robosats/tree/main/scripts/traditional/traditional-services) `strfry_setup` as you did in `nginx_setup`.
Even though this will probably be used to run an external nost relay, so you will not run strfry locally.

alicecoordinator:
Hey @jerryfletcher21. I am not using the traditional scripts.

I saw the loopback IP in strfry.conf, but I decided not to change it, because it would need to be a different value anyway. I mean, if someone wants to run the nostr relay in a different host, the bind in strfry.conf should still be 127.0.0.1, or 0.0.0.0, or some other local interface IP.

I am running my coordinator with a docker compose setup (not the one from robosats-deploy), and in my setup I had configured the nostr relay host with the hostname of the strfry container.

By the way, can you help me with the tests? It looks like they are failing in the main branch.

KoalaSat:
this is great @alicecoordinator can you please also create a PR in https://github.com/RoboSats/robosats-deploy to be compatible with this?

alicecoordinator:
@KoalaSat no changes in robosats-deploy are needed, since localhost will be used as default if the environment variable is not set. :ok_hand:

KoalaSat:
> @KoalaSat no changes in robosats-deploy are needed, since localhost will be used as default if the environment variable is not set. 👌

I understand, but I think it's better to make it explicit to avoid shadowy default values.

alicecoordinator:
@KoalaSat Done: https://github.com/RoboSats/robosats-deploy/pull/39

alicecoordinator:
Thank you!
```
lnbc400u1p5hjq0cpp55suxqad2jd476hzwhwqw6ngpc8ut5eg4gx9zru5c45fh8akqe03sdz2dp68gurn8ghj7emfw3582c3wvdhk6t6jda3x75mpw3ej7un0vfhhxct5wvhhqatvdshnyvehxucqzzsxq97zvuqsp57l444hfdux0c85swuvap3ydjx7c3vwg0w235ckkq2wm7cfqgcgus9qxpqysgquykczzrqgj3h5y93h83h48gvqqk5al5prj4wkjtd5xasdej4y558pksskams07uslusmg0uj936fn7kaqadjt36ytqet4nmgy80pw4sqwhhq6v
```

KoalaSat:
@robosatsdevfund

RobosatsDevFund:
06a296b2a90748420401c584d3715ccaae16d8dac8eb137f9f9d6b9058f6fcf9

PR #2382, opened 2026-01-20
feat: encrypted image upload in chat

### LINKED ISSUE #1206: [feat] Ability to Upload Images in Chat
No

Copy and paste would be fine, but an upload button would also be nice.

**Describe alternatives you've considered**
N/A

**Additional context**
For example - this allows a robot to upload an image of a physical gift card, or a screenshot from an online retailer if there is an issue.

keshav0479:
## What does this PR do?
Fixes #1206

Hi @KoalaSat ,
This PR implements the encrypted image upload feature for the trade chat. As we discussed, I've implemented the full flow using **XChaCha20-Poly1305** for encryption and **BUD-01/02** for Blossom authentication.

### What's Implemented

- **Encryption:** Images are encrypted client-side using XChaCha20-Poly1305 (via `@noble/ciphers`).
- **Authorization:** Uploads to the Blossom server are authenticated using **BUD-01/02 (Kind 24242)** events signed with the robot's nsec.
- **Protocol:**
  - The encrypted blob is uploaded to the coordinator's Blossom server.
  - Decryption keys & metadata are sent to the peer via **NIP-17 Kind 15** file messages.
  - Everything is wrapped in **NIP-59** (Gift Wrap) to ensure privacy and metadata protection.

###  Testing Status 

I've verified the core logic extensively:

- [x] **Encryption/Decryption:** Verified using a standalone test harness (XChaCha20-Poly1305 works perfectly).
- [x] **Blossom Auth:** Verified that the server challenges requests (returns 401) and accepted my auth event structure during dev testing.
- [x]  Code review confirms compliance with NIP-17 and BUD-01 specs.

**Note:** I wasn't able to complete a full "click-to-upload" test in the live chat UI locally because setting up a fully funded Lightning node for the trade flow was a bit of a blocker. The logic is good, but I'd love it if we could verify the final UI integration on a testnet coordinator! 

Looking forward to your feedback!

## Checklist before merging


keshav0479:
> Looking great so far! I;m not sure if I missed it but are we checking the file is an image? I would try to avoid people somehow sending malicious files

Good catch! added a strict check to block anything that isn't an image (like .exe or .sh etc) to be safe.

KoalaSat:
Good news @keshav0479 , our friend @TempleOfSats already have it up and running, in case you want to test it further with a real coordinator: http://ngdk7ocdzmz5kzsysa3om6du7ycj2evxp2f2olfkyq37htx3gllwp2yd.onion/blossom/

keshav0479:
> Good news @keshav0479 , our friend @TempleOfSats already have it up and running, in case you want to test it further with a real coordinator: http://ngdk7ocdzmz5kzsysa3om6du7ycj2evxp2f2olfkyq37htx3gllwp2yd.onion/blossom/

This is awesome, thanks @KoalaSat and @TempleOfSats! i'll give it a spin via Tor and report back.

keshav0479:
Gave it a spin on TempleOfSats, works great!
Verified the full encryption--> upload-->download-->decrypt flow directly against the onion endpoint using a test script. Server accepted the auth and returned the blob correctly.

Screenshot From 2026-01-23 10-34-38

One thing though: i couldn't reach the chat UI locally since opening a trade needs a funded LND node. Is there any way to mock the trade state or jump straight into the chat view for testing? Would love to click the actual button if there's a workaround.

KoalaSat:
> One thing though: i couldn't reach the chat UI locally since opening a trade needs a funded LND node. Is there any way to mock the trade state or jump straight into the chat view for testing? Would love to click the actual button if there's a workaround.

You can use testnet if you have it, if not, then your only option is to actually create a real trade with yourself. But a node is not necessary, just some funds to move forward.

keshav0479:
Very happy to share that i just finished a full end-to-end trade test on TheBigLake - and the image feature is working beautifully! 🎉

Screenshot From 2026-01-23 16-53-48
Screenshot From 2026-01-23 17-12-31
Screenshot From 2026-01-23 17-18-02

Tested & verified:
* PNG, SVG, and 1MB JPEG uploads 
* XChaCha20 encryption/decryption
* Non-image file blocking
One small thing: There's a minor UI glitch where messages can duplicate sometimes (looks like a React state vs Nostr event key conflict). The protocol is solid, just needs a quick dedup fix, which i'll update in the next commit!

keshav0479:
Hey @KoalaSat! I am getting a weird loop where images keep duplicating (+1) every time a new message comes in.
Two things I'm stuck on:
Should I dedup by `sha256` since wrapEvent gives different IDs for self/peer copies?
Any tips on ordering when mixing API chat (int indices) with Nostr (timestamps)? Seeing dropped messages and jumpy order on msgs.
Would appreciate any guidance on this!

KoalaSat:
> Should I dedup by `sha256` since wrapEvent gives different IDs for self/peer copies? 

Can you elaborate?

> Any tips on ordering when mixing API chat (int indices) with Nostr (timestamps)? Seeing dropped messages and jumpy order on msgs. Would appreciate any guidance on this!

If I understood correctly, I think the right approach is to send to the API the url of the image as a message, that way you stay fully out of nostr events when API is selected

keshav0479:
> > Should I dedup by `sha256` since wrapEvent gives different IDs for self/peer copies?
> 
> Can you elaborate?
> 
> > Any tips on ordering when mixing API chat (int indices) with Nostr (timestamps)? Seeing dropped messages and jumpy order on msgs. Would appreciate any guidance on this!
> 
> If I understood correctly, I think the right approach is to send to the API the url of the image as a message, that way you stay fully out of nostr events when API is selected

Ohh that was my bad 😅 ! I was sending images via Nostr even when in API mode and this mixing of two systems caused all chaos.
So the fix: in API mode, upload to Blossom and send the metadata as a PGP message instead of using Kind 15.
One thing i'm unsure about , since the image is encrypted with XChaCha20, the receiver needs the key, nonce, and sha256 to decrypt it. Should i send these as a JSON object in the message body, or is there any other better approach?

KoalaSat:
> So the fix: in API mode, upload to Blossom and send the metadata as a PGP message instead of using Kind 15. One thing i'm unsure about , since the image is encrypted with XChaCha20, the receiver needs the key, nonce, and sha256 to decrypt it. Should i send these as a JSON object in the message body, or is there any other better approach?

We are trying to deprecate PGP so there is no need for long term solutions. I would always use the nostr keys to deal with blossom regardless the network setting

KoalaSat:
One thing to keep in mind, there is the unlikely chance the other robot won´t have a nostr pubkey for using a very old version. Keep that in mind and catch the error.

keshav0479:
Thanks @KoalaSat! Just pushed the updates.
I made sure we're strictly using Nostr keys for Blossom auth in all network modes now. Also caught and fixed that double message bug in API mode (it was sending via both WebSocket and POST, now it's clean).
Tested all flows (Nostr <--> Nostr,  API <---> API etc) and everything is syncing perfectly now, have a look!

Screenshot From 2026-01-26 22-39-24

KoalaSat:
It would be great to warn users (with a Dialog, after clicking the paperclip and before displaying the folder navigator) about uploading files with sensitive data, users should know images are encrypted but they should avoid including sensitive info other than the strictly necessary

KoalaSat:
Images should be able to display bigger when clicking, users will usually use this feature to share QR payments so we should display it as detailed as possible

keshav0479:
Yes that will be great addition for user experience also, on it !
- [x] Privacy Warning Dialog and Image Previews.

keshav0479:
Check these , any modifications ? 

image
image

KoalaSat:
> Check these , any modifications ?

Awesome stuff :rocket:

keshav0479:
Fixed the layout alignment and dual-send logic, and cleaned up that redundant code. It's running smooth across all modes now, have a look!

frontend/src/components/TradeBox/EncryptedChat/index.tsx
```diff
@@ -47,14 +66,132 @@ const EncryptedChat: React.FC<Props> = ({
...
+    if (!nostrPubKey) return;
 
-  const onSendMessage = async (content: string): Promise<object | void> => {
+    const robotNotifications = notifications[nostrPubKey];
+    if (!robotNotifications) return;
+
+    robotNotifications.forEach(([wrappedEvent, unwrappedEvent]) => {
+      handleNostrEventRef.current(wrappedEvent.id, unwrappedEvent);
+    });
+  }, [notifications, garage]);
```
KoalaSat:
```suggestion
  }, [notificationsUpdatedAt, slotUpdatedAt]);
```

frontend/src/utils/blossom.ts
```diff
@@ -0,0 +1,75 @@
...
+  const authToken = createAuthEvent(sha256, nostrSecKey);
+
+  const response = await fetch(uploadUrl, {
+    method: 'PUT',
+    headers: {
+      Authorization: `Nostr ${authToken}`,
+      'Content-Type': 'application/octet-stream',
+    },
+    body: ciphertext.slice().buffer,
+  });
```
KoalaSat: To be compatible with the Android client any external API interaction has been  abstracted, you should use `apiClient.put(`, same applies to any other fetch: https://github.com/RoboSats/robosats/blob/main/frontend/src/services/api/index.ts#L10

KoalaSat (commented):
I think this is my last review, LFG!

keshav0479:
Thanks @KoalaSat ! 🚀 This was a great challenge, really happy to see it merged. Your mentoring made the whole process so much smoother.

Of course! I saw you already accepted my connection on Simplex, feel free to ping me there anytime. Happy to answer anything about the implementation.

```
```

RobosatsDevFund:
062e6681f44367e73e406ad11bba2114c12cb0e9c1640d05328ce0be27474b6e

PR #2383, opened 2026-01-20
Persist book view filters across reloads

### LINKED ISSUE #2372: Store book filters in local storage
When the user reloads the web while working with filters in the book view all selected filters are lost.

 As we do with some settings, we should store these filters in local storage and automatically recover it on load.

We can also take the opportunity to separate the BUY/SELL and Currency book filters from the BUY/SELL  and Currency selector in Maker form, creating different variables. 

It's okay to lost this selection in Maker form on reload.

**Additional context**
Please, consider if there would be any scenario where filters should be automatically clearer out .

GautamSingh0310:
## What does this PR do?
Fixes #2372 

## Checklist before merging
- [ ] utACK

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -141,6 +141,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.0.tgz",
       "integrity": "sha512-UlLAnTPrFdNGoFtbSXwcGFQBtQZJCNjaN6hQNP3UPvuNXT1i82N26KL3dZeIpNalWywr9IuQuncaAfUaS1g6sQ==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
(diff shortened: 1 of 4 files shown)


keshav0479:
Hey @GautamSingh0310 good work on the persistence ,the filters load and save to local storage.
I found one side effect: toggling Buy/Sell or Currency in the "Create Order" form changes the filters on the main Order Book page too. This happens because they share the same global `fav` state.
Hint: To fix this, try uncoupling them by adding `type`, `currency`, and `mode` to the 
`Maker.model.ts` This way the form serves as its own local state and doesn't touch the global book filters until the order is submitted.

GautamSingh0310:
Hey @keshav0479   thanks for the review and hint, i will try it again an update it

KoalaSat:
Thank you all, please @GautamSingh0310 ping me once you are done

PR #2392, opened 2026-01-24
fix: log swap failure reason only once

alicecoordinator:
## What does this PR do?

In the previous implementation, every time the payout_amount method was called, the swap failure reason was logged in the order. This method was called in the order GET method of the API, which led to logging the failure reason dozens of times for some orders.

This commit splits the payout_amount calculation from the context generation. Now the logging happens only in the finalize_contract method, and this can only happen once.

## Checklist before merging


alicecoordinator:
This will prevent these duplicated logs in the order:
image

api/logics.py
```diff
@@ -1380,6 +1370,20 @@ def finalize_contract(cls, take_order):
...
+            )
+
+        if not context["swap_allowed"]:
+            log_message = f"Onchain payment option was not offered: {context['swap_failure_reason']}"
+
+            if config("DISABLE_ONCHAIN", cast=bool, default=True):
+                order.log(log_message)
+            else:
+                order.log(log_message, level="WARN")
+
```
gabbygator184, line 1386: **Failure in `finalize_contract` can leave `TakeOrder` orphaned and cancel an already-LOCKED bond**

`finalize_contract` now calls `compute_buyer_payout_context(order)` (logics.py:1373) between persisting WF2 + the LOCKED `taker_bond` and `take_order.delete()` (line 1387). When `DISABLE_ONCHAIN=False`, this block runs `create_onchain_payment`, which performs gRPC calls to the LN node (`wallet_balance`/`channel_balance`/`estimate_fee`) and can raise. If it does, the `TakeOrder` is never deleted; once it expires (~150s), the sweep in `clean_orders.py:87` calls `take_order_expires` → `cancel_bond(taker_bond)` on an already-LOCKED bond — an invalid LND/CLN operation that can fail. The default (`DISABLE_ONCHAIN=True`) returns early without `create_onchain_payment`, so the risk is limited to coordinators with swaps enabled, but it is still real.

**Suggested fix:** wrap the new block in try/except so `take_order.delete()` and the rest of finalization always run:

```python
try:
    context = Logics.compute_buyer_payout_context(order)
    if "suggested_mining_fee_rate" in context and "swap_fee_rate" in context:
        order.log(
            f"Suggested mining fee is {context['suggested_mining_fee_rate']} Sats/vbyte, the swap fee rate is {context['swap_fee_rate']}%"
        )
    if not context["swap_allowed"]:
        log_message = f"Onchain payment option was not offered: {context['swap_failure_reason']}"
        if config("DISABLE_ONCHAIN", cast=bool, default=True):
            order.log(log_message)
        else:
            order.log(log_message, level="WARN")
except Exception as e:
    order.log(f"Failed to compute buyer payout context: {str(e)}", level="ERROR")
```

PR #2402, opened 2026-01-31
fix: resolve order page loading race condn

### LINKED ISSUE #2400: frontend stuck loading when logging in and switching to order
I noticed often when logging in with a robot token and immediately clicking on a notification or clicking on `Active Order`/`Last Order` the UI will get stuck loading but will never actually load the active order. 
To work around this i then have to click on some other view like the `Offers` and then switch back to my current order so it actually loads.
I recorded this in Tor Browser on linux desktop, but i think it also happens on Android (Tor Browser and in the app). 

https://github.com/user-attachments/assets/ec241b12-faf6-42f1-ac07-e68082e2b7ab

This is what my browser console shows when clicking on the notification popup:

Image
Image

Note: i encountered this on version 0.8.3, however i can reproduce the same issue on 0.8.4 as well

keshav0479:
## What does this PR do?
Fixes #2400

added `slotUpdatedAt` to the OrderPage's dependency array. This ensures the order fetch is automatically retried once the robot's data is fully loaded and auth headers are available.

## Checklist before merging

### DIFF
--- a/frontend/src/basic/OrderPage/index.tsx
+++ b/frontend/src/basic/OrderPage/index.tsx
@@ -25,8 +25,14 @@ import { type UseGarageStoreType, GarageContext } from '../../contexts/GarageCon
 import { genBase62Token } from '../../utils';
 
 const OrderPage = (): React.JSX.Element => {
-  const { windowSize, setOpen, acknowledgedWarning, setAcknowledgedWarning, navbarHeight } =
-    useContext<UseAppStoreType>(AppContext);
+  const {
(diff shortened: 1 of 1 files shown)


keshav0479:
Thank you so much! 

```
```

KoalaSat:
@robosatsdevfund

RobosatsDevFund:
483e6436cca5fca79c8af8304442cc4da4457f170efffe2b68e689323d7f9a7d

PR #2404, opened 2026-01-31
fix: resolve infinite loading and show errors on order renewal

### LINKED ISSUE #2403: recreating order gets stuck
When an order expires and I click on the "recreate(?) order" button in the order view it keeps loading indefinitely instead of leading me to the new bond invoice.
I noticed this with Tor Browser on Android, Robosats 0.8.4.
I suspect the reason could be that the maximum amount of the expired order is higher than what the same coordinator can currently offer, maybe the coordinator rejects it but the frontend doesn't handle this?

https://github.com/user-attachments/assets/dcdcfe30-4a59-4b85-a8c7-eac0bfbe5ba0

Image

keshav0479:
## What does this PR do?
Fixes #2403

resolved infinite loading by properly catching API errors in `renewOrder`
Also added explicit error display to `ExpiredPrompt` because `OrderPage` suppresses global errors for expired orders.

## Checklist before merging

### DIFF
--- a/frontend/src/components/TradeBox/Prompts/Expired.tsx
+++ b/frontend/src/components/TradeBox/Prompts/Expired.tsx
@@ -28,6 +28,13 @@ export const ExpiredPrompt = ({
           {t(order.expiry_message)}
         </Typography>
       </Grid>
+      {order.bad_request && (
+        <Grid item style={{ width: '100%' }}>
+          <Typography variant='body2' align='center' color='error'>
+            {t(order.bad_request)}
+          </Typography>
+        </Grid>
+      )}
       {order.is_maker ? (
         <>
           {order.has_password && (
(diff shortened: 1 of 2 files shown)


KoalaSat (approved):
Great solution and good catch @keshav0479 !

Please accept a small tip for your work. share here a LN invoice for 30,000 sats with a long expiration date

keshav0479:
Thank you so much! :rocket: 

```
```

KoalaSat:
@robosatsdevfund

RobosatsDevFund:
@keshav0479  We tried to pay an LN invoice to the address, but it got stuck. Can you please confirm if you received the payment?

keshav0479:
> @keshav0479  We tried to pay an LN invoice to the address, but it got stuck. Can you please confirm if you received the payment?


just confirmed on my end, and I haven't received yet . It looks like it didn't go through.

PR #2414, opened 2026-02-04
fix: prevent satoshis_now overwrite in OrderView

### LINKED ISSUE #2412: Incorrect amount calculation when taking an order
**Describe the bug**
I was trying to perform an on-chain swap (swapping in some on chains sats for ln sats) and the amount being displayed that I will receive was incorrect

Image

**To Reproduce**
1. Take a swap in order
2. Set the amount
3. When you need to lock your bond then the amount you will receive is not being calculated correctly

**Expected behavior**
It should display the correct amount I'll receive in my buy or swapping in operation

FedericoLuque:
## What does this PR do?
Fixes #2412

This PR fixes a bug in `api/views.py` where the satoshis_now value was being unconditionally overwritten by order.last_satoshis at the end of the OrderView processing.

## Checklist before merging

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -269,9 +269,13 @@ class OrderView(viewsets.ViewSet):
         # 4) If order is between public and WF2
         if order.status >= Order.Status.PUB and order.status < Order.Status.WF2:
             data["price_now"], data["premium_now"] = Logics.price_and_premium_now(order)
-            if take_order.exists():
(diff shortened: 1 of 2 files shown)


KoalaSat:
@FedericoLuque Can you add some test cases to `tests/test_trade_pipeline.py` ?

FedericoLuque:
@KoalaSat I haven't been able to deploy the application locally to visually verify the change; I've had a lot of trouble trying. But the tests I've created are passing.

PR #2418, opened 2026-02-06
fix(api): return 200 with [] for empty /api/book

### LINKED ISSUE #2416: `/api/book` returns HTTP 404 when there are no orders
Hi
While integrating the RoboSats coordinator API, I noticed the following behavior.

```bash
curl -i --socks5-hostname 127.0.0.1:9050 \
  "http://otmoonrndnrddqdlhu6b36heunmbyw3cgvadqo2oqeau3656wfv7fwad.onion/api/book/?format=json&currency=2&type=1"

HTTP/1.1 404 Not Found
Content-Type: application/json

{"not_found":"No orders found, be the first to make one"}
```

Is returning HTTP 404 in the “no orders” case intentional?

From a client perspective, this makes it harder to distinguish between:

- a healthy coordinator with an empty book
- a real endpoint / coordinator error

Would returning 200 OK with an empty array (or a structured response) be acceptable for this case?

Happy to adapt client-side either way, just looking for clarification on the intended semantics.
Thanks!

LouisLau-art:
Fixes #2416.

## Summary
- Change `/api/book` empty-result behavior from `404` with error payload to `200` with `[]`.
- Keep the existing successful response shape unchanged when public orders exist.
- Add a regression test for the empty book case.

## Changes
- `api/views.py`
  - `BookView.get`: return `Response([], status=200)` when no public orders match filters.
- `tests/test_trade_pipeline.py`
  - Add `test_book_empty` to assert `200` and empty list response.

## Validation
- `python -m py_compile api/views.py tests/test_trade_pipeline.py`

I attempted to run the Django test target for `test_book`/`test_book_empty`, but full runtime dependency setup in this environment required long native builds (`grpcio` and others) and could not be completed within this run.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -752,11 +752,8 @@ class BookView(ListAPIView):
                 currency=currency, type=type, status=Order.Status.PUB
             )
 
-        if len(queryset) == 0:
-            return Response(
-                {"not_found": "No orders found, be the first to make one"},
-                status=status.HTTP_404_NOT_FOUND,
-            )
+        if not queryset.exists():
+            return Response([], status=status.HTTP_200_OK)
 
         book_data = []
         for order in queryset:
--- a/frontend/src/models/Coordinator.model.ts
+++ b/frontend/src/models/Coordinator.model.ts
@@ -196,16 +196,13 @@ export class Coordinator {
     apiClient
       .get(this.url, `/api/book/`, undefined, true)
       .then((data) => {
-        if (!data?.not_found) {
-          this.book = (data as PublicOrder[]).reduce<Record<string, PublicOrder>>((book, order) => {
-            order.coordinatorShortAlias = this.shortAlias;
-            return { ...book, [`${this.shortAlias}${order.id}`]: order };
-          }, {});
-          void this.generateAllMakerAvatars();
-          onDataLoad();
-        } else {
-          onDataLoad();
-        }
+        const orders = Array.isArray(data) ? data : [];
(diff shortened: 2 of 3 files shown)


KoalaSat:
This will probably require some changes on frontend also to deal with this new response

LouisLau-art:
Thanks, that makes sense. I can update this PR with the frontend handling for the new empty-book response () so behavior stays consistent end-to-end.

LouisLau-art:
Follow-up: I meant handling the empty-book response shape (HTTP 200 with an empty JSON array), so frontend and backend stay aligned.

LouisLau-art:
Follow-up (formatting fix): I pushed a frontend update in this branch. The book loader now normalizes the API payload to an array before reducing orders, so the empty-book response (HTTP 200 with empty array) is handled explicitly and safely.

juanfradb (approved):
ACK 651800dcebb8d43848be39ceb9367429e8f90968\n\nReviewed diff and tested locally.\n\nValidation run:\n- docker compose test stack with host ports disabled locally due port 8080 conflict\n- docker exec test-coordinator coverage run manage.py test tests.test_trade_pipeline.TradeTest.test_book_empty tests.test_trade_pipeline.TradeTest.test_book\n- docker exec test-coordinator ruff check api/views.py tests/test_trade_pipeline.py\n- npx eslint src/models/Coordinator.model.ts\n\nThe empty-book API response now matches the documented 200-array shape, and the existing non-empty book behavior still passes.

PR #2423, opened 2026-02-10
fix: order logs rendering

alicecoordinator:
## What does this PR do?
Fixes error while rendering order logs.
image

In api/logics.py, the log line was being added without interpolation, so the exact string below was added to the logs:
```
Robot({slashed_robot.id},{slashed_robot.user.username}) was returned {slashed_return} Sats)
```

When this log was rendered, the function `format_html` would try to interpolate values within brackets (`slashed_robot`), and would fail because no argument is being passed.

It would only make sense to use `format_html` if we are passing arguments like this:

```python
# evaluates to <div>Hello &lt;3</div>
format_html('<div>{escaped}</div>', escaped='Hello <3')
```

## Checklist before merging


alicecoordinator:
@jerryfletcher21 I cherry picked the commit from your branch, and rebased my branch. Would you mind submitting an approval just for documentation purposes?

jerryfletcher21 (approved):

KoalaSat (approved):

alicecoordinator:
The fix has already been pushed in https://github.com/RoboSats/robosats/commit/382d7850d1aa018de1ca2fb300153464cd7f3ed1.

Issue #1230, opened 2024-04-17
Buyer stucks on last step while preparing an order


KoalaSat:
**Describe the bug**
When coordinator's confirmed liquidity is bellow the order amount, the buyer got stuck on the last step.

 Reproduce**
1. Your coordinator should have low confirmed on chain liquidity
2. Create and publish a SELL order with robot1 for an amount higher than coordinator's liquidity
3. Take the order with robot2
4. Block robot2's escrow
5. Robot2 should be moved to step 4 where they set the LN invoice (or onchain address) but it got an infinite loading instead

**Expected behavior**
Robot2 should be able to go to the last step.

**Additional context**
The back-end is returning 500 error because `lndcli` command is failing while trying to get either `estimate_fee`or `wallet_balance`, the back-end should expect and control it so the API can print a valid answer.

PR #2429, opened 2026-02-17
fix bug low onchain balance results in infinite loading

jerryfletcher21:
There is a bug that happens when the coordinator has low onchain balance that results in an infinite loading on the frontend when an order bigger than the coordinator onchain balance is taken.
This happens because `create_onchain_payment` returns `False` when `preliminary_amount > available_onchain` before adding `suggested_mining_fee_rate` to `onchain_payment`.
So `suggested_mining_fee_rate` was called without checking if `create_onchain_payment` returned `True`

### DIFF
--- a/api/logics.py
+++ b/api/logics.py
@@ -746,10 +746,11 @@ class Logics:
             valid = cls.create_onchain_payment(
                 order, user, preliminary_amount=context["invoice_amount"]
             )
-            order.log(
-                f"Suggested mining fee is {order.payout_tx.suggested_mining_fee_rate} Sats/vbyte, the swap fee rate is {order.payout_tx.swap_fee_rate}%"
-            )
-            if not valid:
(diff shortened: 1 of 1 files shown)


keshav0479:
I checked this locally too, looks related to the low-balance path from #1230.

When `create_onchain_payment()` returns `False`, `order.payout_tx` is still empty, but current main tries to log `suggested_mining_fee_rate` before checking `valid`, so the backend can 500 instead of returning normal "swap not allowed" response.

So this guard makes sense to me.

One small related thing i noticed while tracing it: the `OnchainPayment` row is created before the balance check, so this path may leave an unattached `CREAT` row behind. Maybe cleanup already handles that, just mentioning since it’s in same flow.

KoalaSat:
Already fixed in a new PR. Thanks for reporting!

PR #2434, opened 2026-02-18
fix: allow null premium default to unblock order creation

### LINKED ISSUE #2433: Bug: "Create Order" button stays disabled when premium field is not manually edited
**Describe the bug**

When creating a new order, if the user fills in all required fields (buy/sell, currency, amount, payment method) but does not manually type anything in the "Premium over Market (%)" field, the "Create Order" button stays greyed out and the order cannot be submitted. The user expects that leaving the premium field untouched would default to 0% (market price), but instead the form treats it as invalid. A tooltip saying "The premium must be a number" appears on hovering the disabled button, but it's easy to miss.

This happens because the default value for premium is set to null internally ```(Maker.model.ts)```, and the form validation requires it to be a number type. Until the user manually clicks into the premium field and triggers a change event, it stays as null and blocks submission.

**To Reproduce**

1.Go to the "Create Order" page on RoboSats
2.Select Buy or Sell
3.Choose a currency and enter an amount
4.Select a payment method
5.Do not touch the "Premium over Market (%)" field — leave it as-is
6.Try to click "Create Order"
7.The button remains greyed out / disabled — order cannot be created

**Expected behavior**
  The "Create Order" button should be enabled and the order should be created with a default premium of 0%                 (market price) when the user has not manually entered a premium value.

**Screenshots**

Image

"All fields filled but Create Order button remains disabled. Premium field is empty (default state). Notice the summary text says '0% discount' but the button is still greyed out."

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
Restarting Tor Browser sometimes fixes it because the user ends up clicking into the premium field, which triggers the handler that converts null to 0. The root cause is that disable Submit in ```MakerForm.tsx``` blocks on ```typeof maker.premium !== 'number'```, which rejects the default null value. The fix is to allow null as a valid premium (treating it as 0%) and use nullish coalescing (?? 0) in price/limit math.

kanishka0411:
## What does this PR do?
Fixes #2433

"Create Order" button stays disabled when the premium field is left untouched, because defaultMaker sets premium: null and the validation rejects non-number types. Fixed by allowing null premium in disableSubmit (treating it as 0% market price) and using ?? 0 in price/limit math to prevent NaN.

## Checklist before merging


alicecoordinator:
I believe it would make sense to restrict this behavior to buy orders. It is less likely that a seller would like to sell for market price, and one could accidentally forget to fill the premium field.

KoalaSat:
I agree with @alicecoordinator not explicitly setting premium should be a hard blocker

kanishka0411:
Fair point, hadn't thought about the seller side. Someone forgetting the field and publishing at market price is a real risk, so I'll make it a hard blocker.
One thing I'd still like to fix though: there's currently no visible reason the button is disabled, just a tooltip that's easy to miss, and the summary underneath still says "at a 0% discount" while it's greyed out. That's what threw me off in the first place. So I'd keep the block but put the error on the premium field itself.

Should I block both buy and sell, or only sells?

KoalaSat (approved):

KoalaSat:
Thanks!

PR #2450, opened 2026-03-15
Log status change only if status really changed

alicecoordinator:
## What does this PR do?
- Log order status change only if status really changed


KoalaSat:
We already have `self.log_status_transition` on that same line, I assume this is an older PR

PR #2479, opened 2026-04-26
fix: apply reward routing budget to payouts

keshav0479:
## What does this PR do?
Related to #1382  
Follow-up to #2038 and #2364

While rebasing #2364, I looked at the reward backend path again.

#2038 added `routing_budget_ppm` to `/api/reward/`, but the reward `LNPayment` was not storing the budget fields. So the invoice amount could be checked with the submitted budget, while the actual LND/CLN reward payout still used the default reward fee cap.

This stores the reward budget on the `LNPayment` and uses the stored `routing_budget_sats` when paying rewards.

I also fixed a failure path where reward payout could return plain `False`, but `withdraw_rewards()` expected `(paid, failure_reason)`. In that case the reward balance had already been cleared, so now it returns a normal invoice error and restores the original reward balance.

Tested:
- `./venv/bin/python manage.py test api.tests.test_logics --verbosity 2`
- `./venv/bin/ruff check api/logics.py api/serializers.py api/lightning/lnd.py api/lightning/cln.py api/tests/test_logics.py`
- `git diff --check upstream/main`

## Checklist before merging

### DIFF
--- a/api/lightning/cln.py
+++ b/api/lightning/cln.py
@@ -494,13 +494,7 @@ class CLNNode:
         """Sends sats. Used for rewards payouts"""
         from api.models import LNPayment
 
-        fee_limit_sat = int(
-            max(
-                lnpayment.num_satoshis
-                * float(config("PROPORTIONAL_ROUTING_FEE_LIMIT")),
-                float(config("MIN_FLAT_ROUTING_FEE_LIMIT_REWARD")),
-            )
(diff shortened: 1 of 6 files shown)


KoalaSat:
Ey @keshav0479 sorry for taking this long but the recent security fixed caused conflicts with this PR

keshav0479:
@KoalaSat updated this one too , now it should be fine

PR #2500, opened 2026-05-25
Assign Alice its own ports, next sequential

stupleb:
## What does this PR do?

In v0.8.5-alpha Alice and Freedomsats are wired to the same local socat ports, and Alice's nginx upstream points at Bazaar's bridge.

robosats-client.sh launches multiple socats but two collide and don't end up listening. Whichever of mainnet_alice_socat / mainnet_freedomsats_socat loses the bind() race exits with EADDRINUSE

The order-book screen on the web client never finishes loading because requests to /mainnet/alice/api/... are proxied to 127.0.0.1:107 which is Bazaar's bridge and surface Bazaar's responses (or 502s) as Alice's.


KoalaSat:
Already fixed in another PR, thank you for reporting!

PR #2510, opened 2026-06-12
fix(frontend): show exact Sats for swap order amounts

21Mill:
## Problem
  
  Swap orders (currency code `1000`) are denominated in Sats, but the order amount in `OrderDetails` was rendered through `amountToString()`, which rounds to **4 significant figures** (`toPrecision(4)`).
  
  For a real order with `amount = 0.01012540 BTC`, the interface showed **`1,013,000 Sats`**, while the **"Confirm sent"** button — which uses the exact `order.amount` via `toFixed(8)` — showed **`0.01012540 BTC = 1,012,540 Sats`**.
  
  The two figures disagree. The peer expects exactly `order.amount` (`1,012,540` Sats), so the rounded value in the interface is misleading and could cause the payer to send the wrong amount.
  
  ## Root cause
  
  `toPrecision(4)` is the right rounding for fiat amounts, but when the swap amount is expressed in Sats (`amount * 1e8`), it collapses `1012540` to `1.013e6`  → `1,013,000`.
  
  ## Fix

  - Add a dedicated `btcToSatsString()` helper in `prettyNumbers.ts` that formats the exact integer Sat amount (with range support), using `Math.round(btc * 1e8)` instead of significant-figure rounding.
  - Use it in `OrderDetails` for the `currency === 1000` branch, so the displayed Sats always match the exact contract amount (and the "Confirm sent" button).
  - The intentionally compact significant-figure rounding in the book table / chart tooltip / maker preview is left untouched.

  ## Tests
  
  Adds unit tests for `btcToSatsString`, including a regression test: `0.01012540 BTC → "1,012,540"` (previously `1,013,000`).
  
  ## Verification
  
  Confirmed against a real order from the backend (`amount: "0.01012540"`, `currency: 1000`, `payment_method: "On-Chain BTC"`): the expected amount is `1,012,540` Sats, which the fix now renders correctly.

## Checklist before merging

PR


KoalaSat:
@21Mill sorry for taking this long, thank you!

KoalaSat (approved):

PR #2511, opened 2026-06-14
Fix reward routing budget payouts

### LINKED ISSUE #1382: Claiming rewards does not support routing_budget_ppm

caydyan:
## Summary

Fixes #1382. Reward withdrawals now carry the selected routing budget through the whole payout path instead of validating against one budget and paying with the default fee cap.

Changes:
- Store `routing_budget_ppm` and `routing_budget_sats` on reward withdrawal `LNPayment` rows.
- Use the stored reward payout routing budget in both LND and CLN reward `pay_invoice()` calls, with legacy fallback for old rows that have no stored budget.
- Restore the original reward balance if a custom-budget reward payment fails, instead of restoring only the reduced invoice amount.
- Extend the unilateral-cancel reward withdrawal test to use a non-zero routing budget and assert the stored payout amount/budget.

## Validation

Passed locally:
- `python3 -m py_compile api/logics.py api/lightning/lnd.py api/lightning/cln.py tests/test_trade_pipeline.py`
- `git diff --check`
- `/tmp/robosats-1382-venv/bin/python -m ruff check api/logics.py api/lightning/lnd.py api/lightning/cln.py tests/test_trade_pipeline.py`

Attempted focused Django test:
- `python3 manage.py test tests.test_trade_pipeline.TradeTest.test_withdraw_reward_after_unilateral_cancel_routing_budget --verbosity=2`

That was blocked before test startup because the local Python environment did not have Django installed. I then created a Python 3.12 venv, but full requirements installation was blocked by `psycopg2==2.9.10` needing `pg_config`, which is not available on this machine.

BTC payout address if this is accepted for the sats reward: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/api/lightning/cln.py
+++ b/api/lightning/cln.py
@@ -494,13 +494,15 @@ class CLNNode:
         """Sends sats. Used for rewards payouts"""
         from api.models import LNPayment
 
-        fee_limit_sat = int(
-            max(
-                lnpayment.num_satoshis
-                * float(config("PROPORTIONAL_ROUTING_FEE_LIMIT")),
-                float(config("MIN_FLAT_ROUTING_FEE_LIMIT_REWARD")),
+        fee_limit_sat = int(lnpayment.routing_budget_sats)
+        if fee_limit_sat == 0 and lnpayment.routing_budget_ppm == 0:
+            fee_limit_sat = int(
+                max(
+                    lnpayment.num_satoshis
+                    * float(config("PROPORTIONAL_ROUTING_FEE_LIMIT")),
+                    float(config("MIN_FLAT_ROUTING_FEE_LIMIT_REWARD")),
+                )
             )
-        )  # 1000 ppm or 2 sats
         timeout_seconds = int(config("REWARDS_TIMEOUT_SECONDS"))
         request = node_pb2.PayRequest(
             bolt11=lnpayment.invoice,
--- a/api/lightning/lnd.py
+++ b/api/lightning/lnd.py
@@ -466,13 +466,15 @@ class LNDNode:
         """Sends sats. Used for rewards payouts"""
         from api.models import LNPayment
 
-        fee_limit_sat = int(
-            max(
-                lnpayment.num_satoshis
-                * float(config("PROPORTIONAL_ROUTING_FEE_LIMIT")),
-                float(config("MIN_FLAT_ROUTING_FEE_LIMIT_REWARD")),
+        fee_limit_sat = int(lnpayment.routing_budget_sats)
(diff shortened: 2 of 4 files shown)


caydyan:
CI note: the Python linter passed on this head, but the coordinator test workflow failed before running any tests.

Failure is in `Download static files Artifact`:

`no downloadable artifacts found (expired)` for `django-main-static` from `frontend-build.yml` run `21478603168` dated 2026-01-29. That artifact expired on 2026-04-29.

So the LND/CLN coordinator tests did not reach Docker orchestration or Django test execution on this PR head. Local validation still passed py_compile, diff check, and targeted ruff; local focused Django test remains blocked by missing `pg_config` for `psycopg2` installation on this machine.

Likely unblock: a maintainer needs to dispatch `Build: Frontend All Bundles` on an upstream RoboSats branch (or update the coordinator test workflow to build/use a fresh `django-main-static` artifact). Runs from forks are skipped by the current `allow_forks: false` artifact lookup, so I am not pushing an unrelated frontend-only change to this Python fix PR.

KoalaSat:
Ey @caydyan  sorry for taking this long but the recent security fixed caused conflicts with this PR

PR #2522, opened 2026-07-05
fix: escape user-controlled data in order.log() and objects_to_hyperlinks()

TempleOfSats:
### 1. api/models/order.py — log() method
- Added escape() around level and event before storing in the logs HTML table row
- Any user-controlled data (order IDs, usernames, addresses, descriptions, etc.) is now neutralized at the point of storage

### 2. api/utils.py — objects_to_hyperlinks() function
- Added escape() on the entire logs string before regex substitution
- Defense-in-depth: even if something slipped through the first escape, injected HTML/JS is rendered as text instead of being executed

## Security Impact

Both fixes prevent XSS attacks via the order logs system.


jerryfletcher21:
The change in `api/utils.py` renders the logs unreadable in the admin panel

gabbygator184 (dismissed):

alicecoordinator:
As @jerryfletcher21 pointed out, this will break the rendering. There are also places in the code that add logs containing HTML tags.

I think we should fix this by converting the logs column into a JSON column, then having the log message containing plain-text only, but with a few special behavior, like the existing `objects_to_hyperlinks`, which would allow some *very-limited* markdown-style formatting.

KoalaSat (dismissed):

KoalaSat:
@TempleOfSats any update on this?

gabbygator184 (dismissed):

api/admin.py
```diff
@@ -130,16 +130,7 @@ class OrderAdmin(AdminChangeLinksMixin, admin.ModelAdmin):
...
-            return format_html("<b>No logs were recorded</b>")
-        with_hyperlinks = objects_to_hyperlinks(obj.logs)
-        try:
-            html_logs = format_html(
-                f'<table style="width: 100%">{with_hyperlinks}</table>'
-            )
-        except Exception as e:
-            html_logs = f"An error occurred while formatting the parsed logs as HTML. Exception {e}"
-        return html_logs
+        return format_html(render_order_logs(obj.logs))
```
alicecoordinator: It make more sense to use `mark_safe` instead of `format_html`. I wrote the reason in this PR: https://github.com/RoboSats/robosats/pull/2423.

alicecoordinator:
Well done!

alicecoordinator (dismissed):

alicecoordinator (dismissed):

alicecoordinator:
I asked AI to write a script to convert the old logs into the new format: https://gist.github.com/alicecoordinator/f96151ff1ca00eacd88abf2f83be0724. Just need to execute it from the root of the application (inside the container if it is the case).

alicecoordinator (dismissed):

KoalaSat:
@alicecoordinator @TempleOfSats Recent security updates caused conflicts, can you solve it?

PR #2524, opened 2026-07-14
fix: correct orders_disputes_started tracking in open_dispute

chandachewe10:
list.append() returns None in Python, so assigning its return value to robot.orders_disputes_started silently set the field to None on every dispute after the first, wiping the robot's full dispute history.

Additionally, orders_disputes_started is declared as a Django CharField with validate_comma_separated_integer_list  it must be a comma separated string, not a Python list. The first dispute branch was also wrong for the same reason, assigning a single element list instead of a string.

Fixed by using string concatenation that matches the field's expected format throughout.


alicecoordinator (approved):

KoalaSat (approved):
Thanks!

PR #2531, opened 2026-07-16
fix: low onchain balance causes infinite loading on frontend

gabbygator184:
## Fix: low onchain balance causes infinite loading on frontend

When the coordinator has low onchain balance, `create_onchain_payment` returns `False` (because `preliminary_amount > available_onchain`). However, the code was reading `order.payout_tx.suggested_mining_fee_rate` in a log line regardless of the return value, causing an error that resulted in infinite loading on the frontend.

This PR moves the mining fee log inside the `if valid:` block so it only executes when the onchain payment was successfully created.


KoalaSat (commented):
Finally! Is it possible to add some test to validate it?

gabbygator184:
@KoalaSat tests added. Let me know if anything needs to be modified.
There was a problem with the checks. I didn't understand the reason.

alicecoordinator (approved):

TempleOfSats (approved):

PR #2560, opened 2026-08-11
Live DevFund donation weights for coordinator ranking

gabbygator184:
## Summary
The DevFund donation percentage used as coordinator lottery weight is now live: coordinators can change their DEVFUND at any time, and the client reads each one's real value to rank coordinators (lottery weight, default MakerForm host, book position).

This also gives new coordinators a path to visibility: since the weight is read live from each coordinator's /api/info/, a newcomer can temporarily raise its DevFund donation — for a few days or weeks — to win the lottery more often, appear as the default MakerForm host and rank higher in the book, attracting more makers while it builds a reputation. The ranking reacts to these changes in real time instead of being locked to the static federation.json snapshot.

## Changes
- Backend: GET /api/info/ now exposes devfund (percentage = DEVFUND × 100). Additive/public.
- Frontend: new services/DevFundProfile.ts reads each coordinator's live devfund by reusing the coordinator's canonical loadInfo() — no dedicated request is issued. A coordinator whose info was already fetched (e.g. by a dialog) is not re-probed; only unreached coordinators are fetched once. Silent requests, 15 s timeout, 30 min module cache. On success it uses the live value; on failure/unreachable it keeps the static federation.json value as fallback.
- Coordinator.loadInfo() now returns a Promise and deduplicates concurrent callers into a single /api/info/ request (infoPromise guard), so overlapping probes never duplicate network calls.
- federationLottery accepts live overrides (clamped to [0, 50]).
- Federation.loadDevFund() re-runs the lottery and reorders coordinators; the default MakerForm host is re-derived from the new top-ranked coordinator.

## Notes
- Clearnet web: only coordinators exposing a clearnet API report live values; Tor-only coordinators keep the static fallback (per-coordinator degradation, mirroring existing fees/book behavior).
- Race guard: the default host is re-derived only once live data arrives and only if the user has not manually chosen a coordinator (coordinatorPicked), so a late shuffle never overwrites an explicit choice.
- Older coordinators that do not yet serve devfund transparently fall back to the static value.

## Testing
- Backend: api/tests/test_info.py (devfund 0/5/20/100%).
- Frontend: ESLint clean on changed files; tsc --noEmit adds 0 new errors (repo baseline has unrelated pre-existing errors).

## Checklist before merging


jerryfletcher21:
I like the idea, just why limiting it to coordinators exposing clearned API?
> Clearnet web: only coordinators exposing a clearnet API report live values; Tor-only coordinators keep the static fallback (per-coordinator degradation, mirroring existing fees/book behavior).

frontend/src/contexts/GarageContext.tsx
```diff
@@ -92,6 +105,13 @@ export const GarageContextProvider = ({
     };
   }, []);
 
+  useEffect(() => {
+    if (!federation.devFundLoaded || coordinatorPickedRef.current) return;
+    setMaker((maker) => {
+      return { ...maker, coordinator: federation.getCoordinatorsAlias()[0] };
+    }); // re-derive default MakerForm coordinator once live devfund data arrives
+  }, [federation.devFundLoaded]);
```
KoalaSat: Why `coordinatorPickedRef`? 

I think it's a bit of over-coding, I understand the intention is to assign the coordinator to the maker form only after all coordinators are loaded, but you can do that directly in the maker form file and just make the coordinator selector unavailable (or loading) in the mid time

frontend/src/models/Federation.model.ts
```diff
@@ -187,6 +189,28 @@ export class Federation {
...
+      });
+
+      const order = federationLottery(defaultFederation, overrides);
+      const reordered: Record<string, Coordinator> = {};
+      order.forEach((alias) => {
+        if (this.coordinators[alias]) reordered[alias] = this.coordinators[alias];
+      });
+      Object.keys(this.coordinators).forEach((alias) => {
+        if (!reordered[alias]) reordered[alias] = this.coordinators[alias];
+      });
```
KoalaSat: I think this code can be simplified

frontend/src/contexts/GarageContext.tsx
gabbygator184: The idea behind `coordinatorPickedRef` was to prevent the live-lottery reorder from overwriting a manually-selected coordinator after mount, but I removed it with this simplification by disabling the selector as you proposed.

frontend/src/models/Federation.model.ts
gabbygator184: The double loop was replaced by a single pass that reorders coordinators by lottery order, preserving the original behavior with less code.

gabbygator184:
> I like the idea, just why limiting it to coordinators exposing clearned API?
> 
> > Clearnet web: only coordinators exposing a clearnet API report live values; Tor-only coordinators keep the static fallback (per-coordinator degradation, mirroring existing fees/book behavior).

This applies only to clearnet web access. Tor-only coordinators are indeed inaccessible via the clearnet browser — a browser cannot reach .onion addresses without a Tor proxy, so they fall back to the static federation.json value.

If a coordinator is accessible AND sending `devFund`, it will be included in the overrides map and used by `federationLottery`.

gabbygator184:
Tests updated and working.

frontend/src/components/MakerForm/MakerForm.tsx
```diff
@@ -68,6 +68,8 @@ const MakerForm = ({
   const { federationUpdatedAt } = useContext<UseAppStoreType>(AppContext);
   const { maker, setMaker, garage } = useContext<UseGarageStoreType>(GarageContext);
 
+  const coordinatorAssignedRef = useRef(false);
```
KoalaSat: I think this is no longer needed

KoalaSat (dismissed):

gabbygator184:
@KoalaSat , I had to uninstall `pre-commit` to be able to commit. I was getting the following error:
```
tsconfig.json:12:25 - error TS5095: Option 'bundler' can only be used when 'module' is set to 'preserve' or to 'es2015' or later.
```
I changed `"module": "commonjs"` to `"module": "esnext"` in `frontend\tsconfig.json`, and several other errors unrelated to my changes appeared. It looks like there’s something wrong with #2624 .

TempleOfSats:
Hey @gabbygator184 , are you comfortable if I add the changes of #2561 here? sorry, I should have done that from the beginning probably.

gabbygator184:
@TempleOfSats , update yours when this one is merged into the main branch. It's better.

TempleOfSats (approved):

KoalaSat:
> I changed `"module": "commonjs"` to `"module": "esnext"` in `frontend\tsconfig.json`, and several other errors unrelated to my changes appeared. It looks like there’s something wrong with #2624 .

Probably you didn´t `npm install` after the upgrade and you are facing version conflicts in your local

KoalaSat (approved):

PR #2561, opened 2026-08-12
Weight lottery by actual sats donated (devfund% × fee rate)

TempleOfSats:
Follow-up to #2560

The lottery weight is now devfund% × (maker_fee + taker_fee) instead of devfund% alone. A coordinator with high devfund% but low fees no longer outranks one that donates more actual sats.

Changes:
- federationLottery.ts: Added feeOverrides parameter and DEFAULT_TOTAL_FEE constant
- Federation.model.ts: Build feeOverrides map from live /api/info/ probe data

Static fallback uses 0.2% default total fee. No backend changes needed.


alicecoordinator:
Good idea to fetch the updated devfund live. I wish we would also move more stuff from the federation.json to endpoints.

alicecoordinator (dismissed):

alicecoordinator:
Only now I read that this is a follow up to https://github.com/RoboSats/robosats/pull/2560. I think both changes are reasonable, so I think we can merge this PR (I guess the other one will be automatically closed).

frontend/src/utils/federationLottery.ts
```diff
@@ -13,10 +13,24 @@
...
+  federation: Record<string, CoordinatorSeed> = defaultFederation,
+  devfundOverrides: Record<string, number> = {},
+  feeOverrides: Record<string, number> = {},
+): string[] {
+  return Object.values(federation)
     .map((coor) => {
-      const chance = coor.badges.donatesToDevFund > 50 ? 50 : coor.badges?.donatesToDevFund || 0;
+      const rawDevfund = devfundOverrides[coor.shortAlias] ?? coor.badges?.donatesToDevFund ?? 0;
+      const devfund = Math.min(50, Math.max(0, rawDevfund));
+      const totalFee = feeOverrides[coor.shortAlias] ?? DEFAULT_TOTAL_FEE;
```
gabbygator184: The lottery should cap `totalFee` the same way it already caps `devfund`. Currently a coordinator can boost its lottery share just by raising fees. But a coordinator that raises fees is already hosting a reasonable volume of orders, so a good portion of users already choose it regardless of the lottery. Doubling the fee, moreover, provides a disproportionate boost to the random ordering of the coordinators.

```
const totalFee = Math.min(
  DEFAULT_TOTAL_FEE,
  Math.max(0, feeOverrides[coor.shortAlias] ?? DEFAULT_TOTAL_FEE),
);
```

KoalaSat (approved):

PR #2614, opened 2026-08-13
Remove coordinator unreachable error banner

TempleOfSats:
Removes the dispatcher error message for coordinator unreachable cases. The error is already thrown and handled by the caller, so the redundant dispatchError call is unnecessary.


KoalaSat (approved):

KoalaSat:
Linterns fixed in https://github.com/RoboSats/robosats/pull/2616

PR #2623, opened 2026-08-15
remove also gpg keys when cleaning users

jerryfletcher21:
remove also gpg keys from the keyring when removing an old unused user


alicecoordinator (approved):

KoalaSat (approved):

KoalaSat:
Can you add some tests before merging?

PR #2633, opened 2026-08-16
docs: fix broken "How to use" links in payment methods

### LINKED ISSUE #2620: Broken "How to use" link
**Describe the bug**
Broken link

**To Reproduce**
1. Go to https://learn.robosats.org/docs/payment-methods/
2. Click on "How to use"
3. Arrive at Github 404 (https://github.com/RoboSats/robosats/blob/main/docs/how-to-use.md)

21Mill:
## What does this PR do?
Fixes #2620

The "How to use" link on the payment methods pages pointed to `docs/how-to-use.md`,
  a file that no longer exists, so it returned a GitHub 404. While fixing it I found
  the three translations were broken as well:

  | Page | Before | After |
  | --- | --- | --- |
  | `01-best-practices/02-payment-methods.md` | `blob/main/docs/how-to-use.md` — 404 (the reported one) | `/read/en/` |
  | `fr/…` | `blob/main/docs/fr/how-to-use.md` — 404 too | `/read/fr/` |
  | `es/…` | right path, wrong org casing (`Robosats`) | `/read/es/` |
  | `pt/…` | linked to the **Spanish** tutorial | `/read/en/` |

  They now point at the rendered site pages rather than GitHub blob URLs. That matters
  beyond the 404: the tutorials reference their screenshots with site-absolute paths
  (`/assets/images/how-to-use/…`), which the blob view resolves against `github.com`,
  so a blob link would land the reader on a page with every screenshot broken. The
  permalinks used here (`/read/en/`, `/read/es/`, `/read/fr/`) are the same ones the
  sidebar navigation already uses. `pt` falls back to `/read/en/` because there is no
  Portuguese tutorial. 

  Also fixed a dead image while in there: the Satoshi's Journal "Robo Garage" header in
  the English and French tutorials (`i0.wp.com/satoshisjournal.com/…/robosats.jpg`)
  returns 404, and so does the linked article — it is gone from the site entirely, not
  just moved. The image is now vendored under `assets/images/how-to-use/` (recovered
  from the Wayback snapshot, 814×458, 96K) and the article link points at its Wayback
  snapshot so the reference stays usable.

  Verified: all 32 local image paths referenced by the tutorials resolve to files in the
  repo, the three target permalinks exist, and no reference to `i0.wp.com` remains.

  Not included: `_data/navigation_pt.yml:100` links to `/read/pt/`, which does not exist
  either — happy to fold that in or leave it for a separate PR.

  ## Checklist before merging
  - [x] Install [pre-commit](https://pre-commit.com/#installation) and initialize it: `pip install pre-commit`, then `pre-commit install`. Pre-commit installs
  git hooks that automatically check the codebase. If pre-commit fails when you commit your changes, please fix the problems it points out.

### DIFF
--- a/docs/_pages/docs/01-best-practices/02-payment-methods.md
+++ b/docs/_pages/docs/01-best-practices/02-payment-methods.md
@@ -20,7 +20,7 @@ Note: This guidance is modified from Bisq's <a href="https://bisq.wiki/Trading_r
 
 ### For both bitcoin buyer and seller
 
-  1. Make sure to checking out the  <a href="https://github.com/RoboSats/robosats/blob/main/docs/how-to-use.md">How to use </a>section before begin trading.<br>
+  1. Make sure to checking out the  <a href="/read/en/">How to use </a>section before begin trading.<br>
   2. State the agreement clearly to prevent misunderstanding.<br>
   3. The fiat payment method should be able to send and receive instantly because the hodl invoice had an expiration time of 24 hours.<br>
 if the timer reaches the expiration, it could trigger a dispute and could lead to a loss of fidelity bond.<br>
--- a/docs/_pages/docs/es/01-best-practices/02-payment-methods.md
+++ b/docs/_pages/docs/es/01-best-practices/02-payment-methods.md
@@ -20,7 +20,7 @@ Nota: Esta guía es una modificación de las reglas de intercambio de   <a href=
 
 ### Para ambos comprador y vendedor de bitcoin
 
-  1. Asegúrate de revisar la seccion <a href="https://github.com/Robosats/robosats/blob/main/docs/_pages/tutorials/read/how-to-use_es.md">Cómo usar </a>antes de empezar a intercambiar.<br>
+  1. Asegúrate de revisar la seccion <a href="/read/es/">Cómo usar </a>antes de empezar a intercambiar.<br>
   2. Indica el acuerdo claramente para evitar malentendidos.<br>
   3. El método de pago fiat debería poder enviar y recibir instantáneamente porque la factura retenida tiene un tiempo de expiración de 24 horas.<br>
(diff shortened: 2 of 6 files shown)


KoalaSat (commented):
Ey @21Mill thanks for the PR! I was thinking maybe we should switch to another available video instead or relaying in the Internet Archive (with services down by the time I write this 😅 

I would say the one still valid and widely used is https://www.youtube.com/watch?v=XW_wzRz_BDI

21Mill:
The Internet Archive thing makes sense.

but the image itself is no longer served from archive.org. I vendored it into `docs/assets/images/how-to-use/` so it loads from learn.robosats.org like every other screenshot in the tutorial. The only remaining Wayback dependency is the `<a>` around the heading, because the article itself is gone too.

The video is already embedded at `_pages/tutorials/watch/videos-en.md:14`, so adding it to /read/en/ would duplicate it across the two tutorial sections. Two options, happy with either:

1. Drop the dead entry. Remove the image + heading + blurb from how-to-use.md  and how-to-use_fr.md, and drop the vendored .jpg. The BTC Sessions video already covers that ground on /watch/en/, one click away in the same sidebar nav.
2. Swap in the video as you suggested, embedded in /read/en/ the same way. Note /watch/fr/ doesn't carry this video, so the French page would end up with an  English-only resource, fine by me if that's acceptable.  

I'd lean towards 1, since it removes the archive.org link entirely without duplicating content. Let me know and I'll push the change.

KoalaSat:
Any of the 2 solutions works for me, feel free to choose :smile:

21Mill:
Went with option 1. Dropped the dead Satoshi's Journal entry from both how-to-use.md and how-to-use_fr.md and removed the vendored image. No archive.org dependency left in the repo. Force-pushed onto the same commit so the binary never lands in history. Thanks!

KoalaSat (approved):
Thank you!

PR #2683, opened 2026-08-26
Federation concensus

KoalaSat:
### The Problem

Today, adding or removing a coordinator from the RoboSats federation requires publishing a new version of every client app. Users have to update before they see the change, and coordinators have to wait for a release cycle to join or leave. This creates unnecessary friction for a system that is supposed to be decentralised and permissionless.

### The Change

After this release, coordinators control their own membership by publishing who they recognise as part of the federation. When enough coordinators agree on a change — a new member joining, or an existing one leaving — every connected client picks it up automatically in the background, without any app update.

### How It Feels to Users

The app continues to work exactly as before on launch — it starts from its own built-in list immediately. Once connected, it quietly consults the coordinators it is already talking to and applies a voting process to determine the agreed-upon list. If the coordinators agree with the client, nothing changes. If they have collectively moved on, the client updates its active list on the fly. Coordinators that were removed are disconnected cleanly; new ones become available right away.

### Trust and Safety

No single coordinator can impose a change unilaterally. Votes are weighted by seniority — coordinators that have been in the federation longer carry more influence than newcomers. A coordinator with at least one year of service is guaranteed a meaningful vote weight, and the oldest coordinators carry the most. A strict majority of total weight must agree before any client adopts a new list. When no majority is reached the client keeps its current list unchanged, so indecision always favours the status quo. This makes it prohibitively expensive to flood the federation with fresh coordinators in order to push through an unwanted change. New coordinators not included in client's default federation will never have a vote.

### Scope

This affects all platforms. The one exception is the self-hosted node app, where new coordinators still require a release to become reachable — but coordinator removals take effect automatically there too.

## Checklist before merging


jerryfletcher21:
I like the idea. How is the algorithm that weights the coordinator by seniority?

api/views.py
```diff
@@ -912,6 +1015,13 @@ def get(self, request):
         except BalanceLog.DoesNotExist:
             context["current_swap_fee_rate"] = 0
 
+        # federation_hash lets clients vote on the federation list using the hash
+        # from each coordinator's /api/info/ response — no separate /api/federation/
+        # request needed in the common case where the list hasn't changed.
+        context["federation_hash"] = _canonical_hash(
```
alicecoordinator, line 1046: Maybe the hash could also be cached?

KoalaSat: It's already cached in 2 places: 

- After being calculated, for 5 min: https://github.com/RoboSats/robosats/pull/2683/changes#diff-7fb412808754eb12ee1b172c34c44eb9ce58b3bba43d62ffcb986ad7917588a8R944
- Together with info response, for 30 secs: https://github.com/RoboSats/robosats/pull/2683/changes#diff-7fb412808754eb12ee1b172c34c44eb9ce58b3bba43d62ffcb986ad7917588a8R1050

KoalaSat:
> How is the algorithm that weights the coordinator by seniority?

Each coordinator gets a vote weight between 1 and 10  based on its age, using two different categories:

- < 1 year old → weight 1–3
- \>= 1 year old → weight 4–10
 
where 10 always goes to the oldest coordinator in that vote round

A file hash wins only if its accumulated weight exceeds 50% of all weight. Ties or fewer than 2 voters → no decision, client keeps its local document.

Ages and consulted coordinators are read only from the client's own trusted data, a coordinator cannot inflate its weight by claiming a fake old date or including new/fake coordinators to its JSON file.

PR #2704, opened 2026-08-30
fix: use ask/bid midpoint for Criptoya exchange rate median

chandachewe10:
The per-exchange median used ask twice, so Criptoya rates were skewed to the ask side instead of the intended bid/ask midpoint.


jerryfletcher21 (approved):

KoalaSat (approved):
Good catch!

PR #2716, opened 2026-09-03
Fix do_accounting crash on disputed swap orders

### LINKED ISSUE #2715: do_accounting permanently breaks the AccountingDay ledger when a swap order is in dispute (order.payout is None)
## Describe the bug

`do_accounting()` crashes when an on-chain **swap** order is in a dispute state.

The problem is that `do_accounting()` assumes every disputed order has a Lightning payout and tries to read:

```python
order.payout.num_satoshis
```

However, swap orders do not use `order.payout`. For an on-chain swap, `Logics.update_address()` creates `order.payout_tx` and sets `order.is_swap = True`.

Because `order.payout` is `None` for these orders, the accounting task fails with:

```text
AttributeError: 'NoneType' object has no attribute 'num_satoshis'
```

The problematic code is in `control/tasks.py`:

```python
if day == today:
    pending_disputes = Order.objects.filter(
        status__in=[Order.Status.DIS, Order.Status.WFR]
    )

    if len(pending_disputes) > 0:
        outstanding_pending_disputes = 0

        for order in pending_disputes:
            outstanding_pending_disputes += order.payout.num_satoshis
```

### Why this causes a bigger problem

This is not only a single task failure. The crash also leaves the accounting data for that day incomplete.

1. `AccountingDay.objects.create(...)` runs before the disputed orders are processed.
2. The crash happens before `accounted_day.save()`.
3. As a result, the `AccountingDay` row remains with its default values for fields such as `net_settled`, `net_paid`, `net_balance`, and the `outstanding_*` fields.
4. On the next run, `do_accounting()` starts from `last_accounted_day + 1 day`, so the damaged day is skipped and never calculated again.
5. Since disputes can remain open in `DIS` or `WFR`, the same problem can happen again every night.

The task runs daily at 23:59, so an unresolved swap dispute can cause incorrect accounting rows for every day until the dispute is resolved.

### To Reproduce

1. Run a coordinator with on-chain swaps enabled (`DISABLE_ONCHAIN=False`).
2. Take an order to `CHA` using an **on-chain address** as the payout method.
3. Confirm that:

   * `order.is_swap == True`
   * `order.payout is None`
   * `order.payout_tx` is populated
4. Open a dispute on the order so it moves to `DIS` or `WFR`.
5. Run the accounting task manually or wait for the scheduled task:

```bash
python manage.py shell -c "from control.tasks import do_accounting; do_accounting()"
```

6. The task fails with:

```text
AttributeError: 'NoneType' object has no attribute 'num_satoshis'
```

7. Check the `AccountingDay` entry in Django admin. The accounting values for that day will remain at their default values.
8. Run the task again on the following days. New incomplete accounting rows will be created, while the previous broken rows will not be recalculated.

### Expected behavior

`do_accounting()` should use the payout method that belongs to the order.

For a swap, it should use `payout_tx`. For a Lightning order, it should use `payout`.

It should also skip an order if no payout method is available instead of stopping the entire accounting task.

For example:

```python
for order in pending_disputes:
    payout = order.payout_tx if order.is_swap else order.payout

    if payout is None:
        continue

    outstanding_pending_disputes += payout.num_satoshis
```

Using `order.is_swap` here is important because `payout_tx` can exist on a non-swap order while a swap is being prepared. The flag tells us which payout method should actually be used.

It would also be useful to make the accounting update more robust so that one unexpected order cannot stop the whole nightly reconciliation.

### Screenshots

N/A — this is a backend Celery task failure and does not have a UI surface.

### Environment

* **OS:** Linux 6.14
* **Python/Django:** `django==5.2.17`
* **Celery:** `celery==5.6.3`
* **PostgreSQL driver:** `psycopg2==2.9.12`
* **Commit:** `94f2fd2c` (`main`)

### Smartphone

N/A — this is a server-side issue and can be triggered from any client.

adity1raut:
## What does this PR do?

Fixes #2715

This PR fixes a bug in `do_accounting()` that could crash when an on-chain swap order was in `DIS` or `WFR` state.

The problem was that `do_accounting()` always tried to get the payout amount from `order.payout.num_satoshis`. That works for Lightning payouts, but not for on-chain swaps. For swaps, `Logics.update_address()` sets `order.payout_tx` and marks the order with `order.is_swap = True`.

Because of this, a disputed swap order could cause:

`AttributeError: 'NoneType' object has no attribute 'num_satoshis'`

There was another issue caused by this crash. The `AccountingDay` row was created before the error happened, so the row could be saved with incomplete accounting values. On the next run, the accounting process would start from the following day, meaning the broken day was never processed again. If the dispute remained open, the accounting job could keep failing every night.

This PR fixes that by choosing the payout amount based on `order.is_swap`:

* For swap orders, use `payout_tx`.
* For Lightning orders, use `payout`.
* If an order does not have a payout method yet, skip it instead of stopping the entire accounting run.

We also use `order.is_swap` rather than checking whether `payout_tx` exists. This is important because `payout_tx` can be created while preparing a swap, even when the order is not actually a swap.

There was also a small issue in the daily disputes calculation. It was subtracting `outstanding_earned_rewards` from the previous day instead of `outstanding_pending_disputes`. This has been corrected as well.

### Tests

Added five regression tests in `control/tests.py`. They cover:

* Lightning payouts
* On-chain swap payouts
* Both possible combinations of `is_swap` and `payout_tx`
* Orders where the payout method is still missing

These tests also reproduce the original `AttributeError` if the helper is changed back to the old `order.payout.num_satoshis` logic.

This closes a coverage gap in the existing tests. `tests/test_trade_pipeline.py` already calls `do_accounting()`, but it did not cover a disputed swap order.

## Checklist before merging

* [x] Install [[pre-commit](https://pre-commit.com/#installation)](https://pre-commit.com/#installation) and initialize it:
  `pip install pre-commit`
  `pre-commit install`

Pre-commit installs git hooks that automatically check the code before committing. If any checks fail, they should be fixed before merging.


adity1raut:
PTAL @KoalaSat

gabbygator184 (approved):

KoalaSat (approved):
Thanks!

PR #2742, opened 2026-09-11
Fix incomplete admin order cancel

gabbygator184:
## What does this PR do?

The `cancel_public_order` admin action was leaving the system in an inconsistent state by failing to expire pending pretaker bonds, missing audit logs, and omitting the required Nostr order book synchronization. This change introduces `Logics.close_public_order` to centralize the cancellation logic (previously duplicated in `cancel_order` and `order_expires`), ensures atomic state transitions, proper TakeOrder cleanup, and robust event notification, aligning admin-initiated cancels with the standard user cancellation flow.

## Checklist before merging

### DIFF
--- a/api/admin.py
+++ b/api/admin.py
@@ -55,8 +55,20 @@ class ETokenAdmin(AdminChangeLinksMixin, TokenAdmin):
 class LNPaymentInline(admin.StackedInline):
     model = LNPayment
     can_delete = True
-    fields = ("payment_hash", "num_satoshis", "status", "routing_budget_sats", "description")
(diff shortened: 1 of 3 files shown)


api/admin.py
```diff
@@ -157,16 +169,27 @@ def cancel_public_order(self, request, queryset):
         """
         for order in queryset:
             if order.status in [Order.Status.PUB, Order.Status.PAU]:
```
KoalaSat, line 162: Given the logic implemented on the function, I'd rather use this statement inside of it. 

It's fine if you prefer to keep this check also here for better code readability.

KoalaSat (approved):

PR #2751, opened 2026-09-14
Fix broken state after onchain admin resolution

KoalaSat:
## What does this PR do?

Fix broken state after onchain admin resolution

## Checklist before merging


api/admin.py
```diff
@@ -305,20 +317,23 @@ def successful_trade(self, request, queryset):
...
-                if order.is_swap:
-                    order.payout_tx.status = OnchainPayment.Status.VALID
-                    order.payout_tx.save(update_fields=["status"])
-                    order.update_status(Order.Status.SUC)
-                else:
+                if not order.is_swap:
                     order.update_status(Order.Status.PAY)
 
-                Logics.pay_buyer(order)
+                paid = Logics.pay_buyer(order)
```
gabbygator184: Swap orders stay stuck in DIS/WFR after resolution. When is_swap, no status transition happens before pay_buyer(). Its swap path calls complete_order(), whose transition_status(SUC) only accepts [FSE, PAY, FAI], so the transition fails.

Fix: always update_status(PAY) before pay_buyer().

Issue #2516, opened 2026-07-01
Worth automating the locale sync?


hiciefte:
I cloned `main` and counted how stale the locale files are against `en.json`. They all still have the full set of keys, but a good chunk is left in English. Catalan is the worst at ~76%, most languages sit around 84–86%, and even the most complete one (Czech) still has ~70 English strings. You can see it in `de.json` with things like `Claim Rewards`, `Swap In`/`Swap Out`, and `You need to enable nostr to receive notifications.`

They're translated by hand, and translations aren't in the reward pool, so I don't think that backlog closes on its own.

I help maintain Bisq's localization tooling ([localize-pipeline](https://github.com/bisq-network/localize-pipeline)), and I could set up a workflow that picks up the changed keys whenever `en.json` changes and opens a PR for review. It stays off until someone adds a model key, it never merges by itself, and it can run against a local model so nothing leaves your servers. It already handles the i18next `{{…}}` variables and leaves the `#N` marker keys alone.

Before I build anything: is this something you'd want? If you'd rather not have AI-drafted translations in the repo at all, that's completely fair — say so and I'll drop it.

cc @Reckless-Satoshi

KoalaSat:
Great idea, we have been looking to implement something like this for a while

hiciefte:
Great to hear, @KoalaSat — thanks for the quick response.

The RoboSats adaptation is already built, so rather than describe it further I'll open a PR you can look at. It adds three files and nothing else:

- `.github/workflows/translate.yml` — runs when `frontend/static/locales/en.json` changes on `main` (also manually dispatchable for a full backfill of the existing gap)
- `translations/config.yaml` — tuned to your flat one-file-per-locale JSON layout
- `translations/glossary.json` — enforced per-language term mappings, empty to start

RoboSats-specific bits handled: i18next `{{placeholders}}` are protected and parity-checked so a locale can't silently drop `{{orderID}}`; your `#N` section-marker keys are preserved verbatim, not translated; all 16 locales from `SelectLanguage.tsx` are covered; and RoboSats, sats, Lightning, Tor, Nostr, PGP etc. are pinned as never-translate. There's also a safety gate that blocks the PR if too many strings come back identical to English or any validation check warns.

[...]

PR #2755, opened 2026-09-15
Add a translation workflow for frontend locales

hiciefte:
# Add a translation workflow for frontend locales

I've put together a starter for keeping the frontend translations up to date.
It's four files: a workflow, config for the existing 16 locales, a starting
glossary, and an operator README. Translation changes come back as PRs to review.

You asked for local inference in
[#2516](https://github.com/RoboSats/robosats/issues/2516). I've configured this
version for hosted OpenAI through AISuite: `gpt-4o-mini` translates and
`gpt-5.6-terra` does a second pass. I wanted the first setup to work with a secret
without needing a model server running first. If you'd rather start local, I can
change it before merge; we'd still need to test that setup. There's no automatic
fallback between providers.

## Before switching it on

With no `OPENAI_API_KEY`, the workflow exits before checkout or model calls.
Once a key is configured, changes to `en.json` on upstream `main` can trigger it.
It also supports manual dispatch. Both actions are pinned to commit SHAs.

I'd start with a manual run, keeping `process_all_files=true` and the default
`locale_filter_glob=es.json`. **Both `*.json` and a blank filter select all 16
locales.** The estimate for a full backfill is roughly 1,900 model calls.
Five concurrent calls and 60 calls/minute limit throughput, not the bill, so set
usage limits with the provider first. Calls can still cost money if opening the
PR fails afterwards.

The config includes text that still matches English. Without a saved ledger,
those entries can be retried on later source pushes. After an accepted backfill,
I'd set `retranslate_identical_source_strings: false` to stop paying for that
work repeatedly.

Actions needs permission to create PRs. Generated commits use the normal GitHub
token and are unsigned; a signing requirement would need extra setup. Check
that CI runs on those PRs, since the normal token may not trigger other
workflows. Each attempt uses a fresh branch, and merging stays manual.

## Terminology

The glossary starts with 93 mappings from existing RoboSats wording across all
16 locales. Of those, 31 also come from our Bisq 2 glossary: shared terms such
as fees, wallets, and trades where the two projects already agree. Where they
differ, I've kept RoboSats' wording, including Spanish `fee → Comisión` and
Italian `trade → Scambio`.

Each entry has examples in at least two RoboSats messages. These are suggestions
for translation and review, with grammatical variations allowed; they aren't
exact-word rules. Some terms still need a translator's decision and are left
out. The README explains the sources and gaps. No existing translations change
in this PR.

## What I've tested

The live test was small: 12 public source entries, nine Spanish translations
that passed validation, and 11 model calls. It preserved an existing translation
and the interpolation counts. It also caught a token-only entry being wrongly
flagged as untranslated. I added a narrow exclusion for that; the repeat made
no model calls and changed nothing.

The 17 offline checks cover the guards, config, locale discovery, JSON
round-trips, excluded keys, interpolation handling, glossary evidence, and
grammatical variation. The new checker also detects the
existing numeric-tag damage in all seven affected locales. The pinned revision is
`385a75b25931e1031f7840d7309632e7f98de26e`. I haven't run a full backfill or this
workflow in GitHub Actions. The Spanish output still needs native-speaker review.

## What still needs human eyes

This pins v0.1.21, which includes the numeric Trans-tag fix from
[pipeline PR #178](https://github.com/bisq-network/localize-pipeline/pull/178).
Tags such as `<1>` and `</1>` are now protected during translation and checked
for matching token counts, alongside `{{var}}` interpolations. That catches
missing or changed tags; it doesn't check component nesting or rendered output.

There is still existing tag damage in `es`, `fr`, `pt`, `zh-SI`, `eu`, `it`, and
`ja`. The affected messages are `unsafe_alert`, `let_us_know_hot_to_improve`, and
`open_dispute`; Portuguese has lost every tag in `unsafe_alert`. Those strings
need separate corrections. Updating the workflow doesn't repair them on its own.

I also found three interpolation-name errors in Catalan, Basque, and French.
Those have a separate small patch; I haven't included it here.

All locale files already have the full set of keys. Much of the backlog is
text still identical to English, and some of that is intentional. Results that
just repeat English are reverted and reported. An individual warning doesn't
always block a PR: the configured thresholds and reported placeholder,
skipped-file, and model failures determine that.

## Guardian

A lot of the work behind this went into
[Localize Guardian](https://github.com/bisq-network/localize-pipeline/blob/385a75b25931e1031f7840d7309632e7f98de26e/docs/guardian.md).
I want a reviewer's correction to carry through to later runs: terminology in
the glossary, missing context in the instructions, and pipeline bugs fixed with
a test so they don't keep coming back.

Guardian reads feedback from allowlisted reviewers, assesses it, and can propose
translation corrections and PRs for pipeline fixes. Those still need someone to
review and follow through on them. It never merges.

It's part of the pinned pipeline, but this workflow doesn't run it or require
it. If you want it later, it's self-hosted: your machine, credentials, and model
budget. The remediation path has only been exercised on `.properties` files,
so I'd want the JSON path verified before letting it touch RoboSats.


translations/README.md
```diff
@@ -0,0 +1,181 @@
...
+1. Operate an OpenAI-compatible endpoint (for example Ollama/vLLM) reachable from
+   the runner. A GitHub-hosted runner's localhost is not your machine. A private
+   endpoint may require a self-hosted runner/network change; do not expose an
+   unauthenticated model server publicly to make this work.
+2. Set `model_name` and `review_model_name` to IDs served by that endpoint, and
+   verify its supported request options and translation quality. AISuite can
+   route its OpenAI provider to a custom endpoint; no provider fallback is used.
+3. Set repository **variable** `LOCALIZE_API_BASE_URL` to its base URL including
+   `/v1`. The guard enables without a key and passes this through `api-base-url`
+   to `OPENAI_BASE_URL`, which overrides YAML `api_base_url`. A non-secret URL
```
KoalaSat, line 162: Can we add `OPENAI_BASE_URL` to the workflow yaml by default? that way we can switch to other compatible provider more respectful with privacy

hiciefte: Done in 1f75edb. `OPENAI_BASE_URL` is now a job-level `env` read from a repository variable of the same name. Empty keeps api.openai.com; set it to any OpenAI-compatible base URL (with `/v1`) and put that provider's token in `OPENAI_API_KEY`. The README notes that a GitHub-hosted runner can't reach localhost, so local inference needs a self-hosted runner or a reachable endpoint. If you'd rather have the URL hard-coded in the YAML so changes go through PR review, that's a one-line switch.

The new push is waiting on workflow approval before CI runs — could you approve it?

PR #2756, opened 2026-09-15
Better coordinator selector

KoalaSat:
## What does this PR do?

Better coordinator selector

## Checklist before merging


gabbygator184:
This PR removed the defensive fallbacks present in the old code. Before: `(coordinator?.info?.maker_fee ?? 0)` and `(coordinator?.info?.current_swap_fee_rate ?? 0)`. Now `CoordinatorFeeRow` directly accesses `info.maker_fee * 100` and `info.current_swap_fee_rate.toFixed(1)`. Since the `/api/info/` response is assigned with a blind cast (`this.info = data as Info`), with no runtime validation, a broken/malicious coordinator replying with `swap_enabled: true` and no numeric `current_swap_fee_rate` triggers a `TypeError` during render — the root `ErrorBoundary` crashes the whole app, and the crash repeats after reload while that coordinator stays enabled (persistent DoS of order creation). I suggest restoring the fallbacks (`?? 0`) and/or sanitizing the numeric fields in `Coordinator.loadInfo()`.

KoalaSat:
> This PR removed the defensive fallbacks present in the old code. Before: `(coordinator?.info?.maker_fee ?? 0)` and `(coordinator?.info?.current_swap_fee_rate ?? 0)`. Now `CoordinatorFeeRow` directly accesses `info.maker_fee * 100` and `info.current_swap_fee_rate.toFixed(1)`. Since the `/api/info/` response is assigned with a blind cast (`this.info = data as Info`), with no runtime validation, a broken/malicious coordinator replying with `swap_enabled: true` and no numeric `current_swap_fee_rate` triggers a `TypeError` during render — the root `ErrorBoundary` crashes the whole app, and the crash repeats after reload while that coordinator stays enabled (persistent DoS of order creation). I suggest restoring the fallbacks (`?? 0`) and/or sanitizing the numeric fields in `Coordinator.loadInfo()`.

Two points from the review don't hold up against the code:

__`maker_fee * 100` and `taker_fee * 100` don't crash.__ Multiplying `null/undefined` by a number yields `NaN`, and `NaN.toFixed(3)` returns the string `"NaN"` — no exception is thrown, React renders `"NaN%"`, which is a display glitch at worst. No fix needed there.

__The `?? 0` fallbacks on `maker_fee`/`taker_fee` weren't "removed defensively" — they were made redundant.__ In the old code `coordinator` itself could be `undefined` (it was a `useState<Coordinator>()`), so `coordinator?.info?.maker_fee ?? 0` was a real guard. In the new `CoordinatorFeeRow`, the explicit `if (!info) return <></>` on line 42 already narrows `info` to `Info` before any field is touched, so `?? 0` on those two fields buys nothing.

gabbygator184 (approved):

PR #2763, opened 2026-09-18
DevFund community split

KoalaSat:
## What does this PR do?

DevFund community split

## Checklist before merging


api/utils.py
```diff
@@ -182,6 +182,119 @@ def get_devfund_pubkey() -> str:
...
+        raise ValueError(
+            f"No invoice (pr) in LNURL-pay callback response for {address}"
+        )
+
+    # --- 5. verify invoice amount ---
+    try:
+        decoded = LNNode.decode_payreq(invoice)
+        invoice_sats = int(decoded.num_satoshis)
+    except Exception as e:
+        raise ValueError(f"Failed to decode invoice from {address}: {e}")
```
gabbygator184, line 287: `LNNode.decode_payreq` returns LND's gRPC `PayReq` (which has `num_satoshis`) on LND, but CLN's `node_pb2.DecodeResponse` has no such field — the amount lives in `amount_msat` (msat). On CLN coordinators `decoded.num_satoshis` raises `AttributeError`, which the broad `except` below converts to `ValueError("Failed to decode invoice...")` — so `send_community_donation` always fails (logged + skipped) on CLN. The feature silently never pays.

gabbygator184 (dismissed):

PR #2765, opened 2026-09-18
Document restricted LND macaroon

ghost:
## What does this PR do?

Documents how coordinator operators using LND can create and use a restricted, least-privilege macaroon instead of giving RoboSats access to the full `admin.macaroon`.

The documented macaroon grants access only to the 14 RPC methods currently used by the RoboSats LND integration. The RPC list was cross-checked against `api/lightning/lnd.py` and tested end-to-end on testnet4 with coordinator order flows.

The `.env-sample` comment is also updated to point coordinator operators to the restricted macaroon documentation.

## Checklist before merging


KoalaSat (approved):
Awesome!

